PR #3, opened 2019-11-21
Integ Test: Run clean before the validation and use the `bash` shell

hossain-rayhan:

### DIFF
--- a/integ/integ.sh
+++ b/integ/integ.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 export AWS_REGION="us-west-2"
 export PROJECT_ROOT="$(pwd)"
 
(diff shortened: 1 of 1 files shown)


integ/integ.sh
```diff
@@ -32,7 +33,10 @@ test_kinesis() {
 	docker-compose --file ./integ/test_kinesis/docker-compose.test.yml build
 	docker-compose --file ./integ/test_kinesis/docker-compose.test.yml up --abort-on-container-exit
 
+    # Giving a pause before running the validation test
+    # Firehose delivery stream has a buffer time which causes the delay to send the data
```
PettitWesley: Kinesis stream

hossain-rayhan: I believe the bffer time and buffer memory are for Kinesis Firehose not for the Kinesis Data Stream.

PR #9, opened 2019-11-24
Remove un-used binaries from repo

PettitWesley:


hossain-rayhan:
Do we have a need to upate the .gitignore?

PettitWesley:
@hossain-rayhan We don't. These binaries won't be created again. They're left over from when I started this project.

PR #13, opened 2019-12-06
Update Cloudwatch integ test exit logic

hossain-rayhan:
*Issue #, if available:*

*Description of changes:*
Though the cloudwatch integ test passes successfully, it is throwing an exception- file not found. It should exit with code zero. 

### DIFF
--- a/integ/integ.sh
+++ b/integ/integ.sh
@@ -9,9 +9,20 @@ test_cloudwatch() {
 	docker-compose --file ./integ/test_cloudwatch/docker-compose.test.yml build
 	docker-compose --file ./integ/test_cloudwatch/docker-compose.test.yml up --abort-on-container-exit
 	sleep 10
+
+	# Creates a file as a flag for the validation failure
+	mkdir -p ./integ/out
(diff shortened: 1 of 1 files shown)


hossain-rayhan:
I ran the integ test and it worked.

Screen Shot 2019-12-06 at 3 00 44 PM

integ/integ.sh
```diff
@@ -56,15 +67,8 @@ clean_s3() {
 }
 
 if [ "${1}" = "cloudwatch" ]; then
-	mkdir -p ./integ/out
-	touch ./integ/out/cloudwatch-test
 	test_cloudwatch
 	clean_cloudwatch
-	if [ -f ./integ/out/cloudwatch-test ]; then
```
PettitWesley: Can the check remain after the call to `clean_cloudwatch`?

Because this change means that now if the test fails then clean does not run

PettitWesley:
@hossain-rayhan I think you need to add the check for the `cicd` target as well

(I'm pretty sure this was what was broken before- the normal make integ target has been working fine)

Issue #11, opened 2019-11-27
Exit Code 137 for newest release


moritz31:
We are triggering fargate tasks with a step function. For the newest fluentbit container this fails because the container always exits with exit code 137 when the actual container in the task finishs.
For the previous version 1.3.2 this is not a problem.

PettitWesley:
@moritz31 I've reproduced this... it is very strange- 137 usually means Out of Memory AFAIK.

PettitWesley:
I think I've got a fix for this.

One key thing to note- the old behavior (afaict) was that the container would always with code 0 even if Fluent Bit exited with a non-zero code. 

@moritz31 Is that what you've seen? My fix will restore that old behavior- the container will (afaict) always exit 0.

PettitWesley:
I think I finally understand this.

137 means that the container did not gracefully handle the SIGTERM, and docker had to SIGKILL it: http://tldp.org/LDP/abs/html/exitcodes.html

Normally when Fluent Bit receives a SIGTERM, it prints a warning that it will quit soon- it has SIGTERM handling built in. I checked my logs and I noticed that with 2.0.0 this does not happen. This means that something about how I set up the entry point script prevents the signals from properly being passed to it. 

The `exec -c` I added in #14 fixes the signal passing somehow.

allisaurus:
Fix released in [v2.1.0](https://github.com/aws/aws-for-fluent-bit/releases/tag/v2.1.0)

PR #14, opened 2019-12-07
Bug fix: container exits with code 0 when sent a SIGTERM

PettitWesley:
*Issue #, if available:*
#11 

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -3,4 +3,4 @@ tput bold;
 echo -n "AWS for Fluent Bit Container Image Version "
 cat /AWS_FOR_FLUENT_BIT_VERSION
 tput sgr0;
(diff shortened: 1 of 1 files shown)


PettitWesley:
This change should return us to the old exit code behavior as detailed in my comment on #11 

I tested the following:
- Fluent Bit encounters an error and exits - leads to code `0` in both 1.3.2 and this change.
- Fluent Bit receives a SIGTERM and gracefully shuts down - leads to code `0` in both 1.3.2 and this change.
- Fluent Bit recieves a SIGKILL because it did not gracefully shut down - leads to code `0` in both 1.3.2 and this change.

PettitWesley:
137 probably comes from not handling SIGTERM properly: http://tldp.org/LDP/abs/html/exitcodes.html

PR #15, opened 2019-12-12
Release 2.1.0

PettitWesley:


Dockerfile
```diff
@@ -3,8 +3,8 @@ FROM amazonlinux:latest as builder
 # Fluent Bit version; update these for each release
 ENV FLB_MAJOR 1
 ENV FLB_MINOR 3
```
hossain-rayhan: Do we still need these three variables- FLB_MAJOR, FLB_MINOR, and FLB_PATCH?

hossain-rayhan:
Before merging this change, we should update the versions and and changelogs for our three plugins.

PR #19, opened 2020-01-08
Publish and Verify image using CI/CD Pipeline

hossain-rayhan:

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -15,8 +15,7 @@ phases:
       - docker images
 
       # Push the image to ECR in the same account and same region the pipeline is hosted.
-      # `amazon/aws-for-fluent-bit` is the image name, and `latest` is the tag name.
-      - ecs-cli push amazon/aws-for-fluent-bit:latest
+      - ecs-cli push amazon/aws-for-fluent-bit-test:latest
 artifacts:
   files:
     - '**/*'
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -1,0 +1,22 @@
+version: 0.2         
+phases:
+  install:
+    runtime-versions:
+      golang: 1.12
+  pre_build:
+    commands:
+      - echo Building the AWS for Fluent Bit image
+  build:
+    commands:
+      # Command to build your project
+      - make release
+
+      # List the docker images 
+      - docker images
+
+      # Push the image to ECR in the same account and same region the pipeline is hosted.
+      # `amazon/aws-for-fluent-bit` is the image name, and `latest` is the tag name.
+      - ecs-cli push amazon/aws-for-fluent-bit:latest
+artifacts:
+  files:
+    - '**/*'
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -15,11 +15,7 @@ phases:
       - 'export AWS_SESSION_TOKEN=`echo $CREDS | jq -r .Token`'
 
       # Pull the image that we built and pushed in the `Build` stage
-      - 'ecs-cli pull amazon/aws-for-fluent-bit:latest'
-                                                               
-      # When we pull image from ECR, it comes with accountId, ECR registry, and region name.
-      # Tag the image only with the name which will be used by the integration test.
-      # Don't forget to update your account and region in the image registry
+      - 'ecs-cli pull amazon/aws-for-fluent-bit:latest'                                                            
       - 'docker tag ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:latest'
 
       # List the images to do a double check
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -15,8 +15,8 @@ phases:
       - 'export AWS_SESSION_TOKEN=`echo $CREDS | jq -r .Token`'
 
       # Pull the image that we built and pushed in the `Build` stage
-      - 'ecs-cli pull amazon/aws-for-fluent-bit:latest'                                                            
-      - 'docker tag ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:latest'
+      - 'ecs-cli pull amazon/aws-for-fluent-bit-test:latest'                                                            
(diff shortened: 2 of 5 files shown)


buildspec.yml
```diff
@@ -0,0 +1,22 @@
...
+  pre_build:
+    commands:
+      - echo Building the AWS for Fluent Bit image
+  build:
+    commands:
+      # Command to build your project
+      - make release
+
+      # List the docker images 
+      - docker images
```
PettitWesley: why list the images?

buildspec_integ.yml
```diff
@@ -0,0 +1,32 @@
...
+  pre_build:
+    commands:
+      - echo Running the integration test
+  build:
+    commands:
+      # Get the default credentials and set as environment variables
+      - 'CREDS=`curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`'
+      - 'export AWS_ACCESS_KEY_ID=`echo $CREDS | jq -r .AccessKeyId`'
+      - 'export AWS_SECRET_ACCESS_KEY=`echo $CREDS | jq -r .SecretAccessKey`'
+      - 'export AWS_SESSION_TOKEN=`echo $CREDS | jq -r .Token`'
```
PettitWesley: I think I might've been the one who originally suggested this... but I now realize that it shouldn't be necessary. The SDK/AWS CLI will query `169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` for you. You should not need to set these env vars.

buildspec_publish_dockerhub.yml
```diff
@@ -0,0 +1,32 @@
...
+    commands:
+      - echo Publish the image to DockerHub
+  build:
+    commands:
+      # Pull the image that we built and pushed in the `Build` stage
+      - 'ecs-cli pull amazon/aws-for-fluent-bit:latest'
+                                                               
+      # When we pull an image from ECR, it comes with accountId, ECR registry, and region name.
+      # Tag the image only with the name.
+      # Don't forget to update your account and region in the image registry.
```
PettitWesley: I don't understand this comment

buildspec_publish_dockerhub.yml
```diff
@@ -0,0 +1,32 @@
...
+      - 'docker tag ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:latest'
+
+      # List the docker images 
+      - docker images
+
+      # DRY RUN: Push the image to dockerhub
+      - './scripts/publish.sh cicd-publish dockerhub'
```
PettitWesley: dry run isn't really needed in CI/CD

The point of dry run is for a human to be able to check that they have permission to log into docker hub.

buildspec_publish_ecr.yml
```diff
@@ -0,0 +1,52 @@
...
+      # Pull the image that we built and pushed in the `Build` stage
+      - 'ecs-cli pull amazon/aws-for-fluent-bit:latest'
+                                                               
+      # When we pull an image from ECR, it comes with accountId, ECR registry, and region name.
+      # Tag the image only with the name.
+      # Don't forget to update your account and region in the image registry.
+      - 'docker tag ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:latest'
+
+      # List the docker images 
+      - docker images
```
PettitWesley: why tho

scripts/publish.sh
```diff
@@ -313,3 +359,34 @@ if [ "${1}" = "rollback-ssm" ]; then
...
 fi
+
+# Publish using CI/CD pipeline
+# Following scripts will be called only from the CI/CD pipeline
+if [ "${1}" = "cicd-publish" ]; then
+	if [ "${2}" = "dockerhub" ]; then
+		publish_to_docker_hub amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_VERSION}
+	elif [ "${2}" = "aws-us-gov" ]; then
+		for region in ${gov_regions}; do
+			gov_cloud_sync ${region} ${gov_regions_account_id}
```
PettitWesley: `gov_cloud_sync` isn't in this PR right?

hossain-rayhan: Yes, you are right. As we talked offline, I will send the gov cloud changes in the next PR. In this PR, this block will never be executed.

buildspec.yml
hossain-rayhan: If anything bad happens, it will help us to debug seeing the logs in our codebuild project. I put it intentionally to see what images we are getting after the build.

buildspec_publish_dockerhub.yml
hossain-rayhan: Yah. Earlier we had hardcoded account id and region. I updated the codes but forgot to remove the comment. Good catch.

buildspec_publish_dockerhub.yml
hossain-rayhan: I was not sure. Thanks. Will remove this step.

buildspec_publish_ecr.yml
hossain-rayhan: Same. Intentionally added for better debugging.

buildspec_integ.yml
hossain-rayhan: I think its necessary for our integ test. Integ test is running in a separate container which I believe somhow search for credentials only in ENV variables. If we do not set these variables, our integ test for cloudwatch fails. See the following logs. 

Successfully tagged test_cloudwatch_logger-log-key:latest 
The AWS_ACCESS_KEY_ID variable is not set. Defaulting to a blank string. 
The AWS_SECRET_ACCESS_KEY variable is not set. Defaulting to a blank string. 
The AWS_SESSION_TOKEN variable is not set. Defaulting to a blank string. 
Creating network "test_cloudwatch_default" with the default driver

PettitWesley: Oh yeah... sorry I forgot about that.

buildspec.yml
```diff
@@ -0,0 +1,22 @@
...
+
+      # Push the image to ECR in the same account and same region the pipeline is hosted.
+      # `amazon/aws-for-fluent-bit` is the image name, and `latest` is the tag name.
+      - ecs-cli push amazon/aws-for-fluent-bit:latest
```
PettitWesley: this will create an ECR repo named "{account ID}.ecr.{region}.amazonaws.com/amazon/aws-for-fluent-bit"

This is really similar to the name of the public repo... and the pipeline runs in the same account that hosts the images. 

Can this temporary repo used by the pipeline have a different name that clearly indicates its purpose?

PR #22, opened 2020-01-15
CI/CD Pipeline: Add sync mechanism and publish SSM parameters

hossain-rayhan:

### DIFF
--- a/buildspec_publish_ssm.yml
+++ b/buildspec_publish_ssm.yml
@@ -10,12 +10,14 @@ phases:
     commands:
       - './scripts/publish_test.sh cicd-publish-ssm ${AWS_REGION}'
 
-      # Assume role to verify, get the credentials, and set them as environment variables
+      # Assume role to verify, get the credentials, and set them as environment variables.
+      # Verification should be done using the credentials from a different account. It ensures that
+      # the ssm parameters we published are public and accessible from any account.
       - CREDS=`aws sts assume-role --role-arn ${VERIFY_ROLE_ARN} --role-session-name ${AWS_REGION} --region ${AWS_REGION}`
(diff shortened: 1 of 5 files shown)


buildspec_publish_ssm.yml
```diff
@@ -0,0 +1,22 @@
...
+    runtime-versions:
+      golang: 1.12
+  pre_build:
+    commands:
+      - echo Publish SSM parameters
+  build:
+    commands:
+      - './scripts/publish_test.sh cicd-publish-ssm ${AWS_REGION}'
+
+      # Assume role to verify, get the credentials, and set them as environment variables
```
PettitWesley: Nit: this comment explains the code, but the code is pretty obvious and readable. Instead, I'd prefer a comment explaining why the code exists, like "Verify using a role in a different account, parameters should be public"

scripts/publish.sh
```diff
@@ -123,6 +123,34 @@ check_parameter() {
...
+		sha2='repo_not_found'
+	fi
+
+	docker images
+	match_two_sha $sha1 $sha2
+
+	if [ "$IMAGE_SHA_MATCHED" = "FALSE" ]; then
+		publish_ecr ${region} ${account_id}
+		publish_ssm ${region} ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit
+	fi
```
PettitWesley: As we discussed yesterday, there's a slight problem with this.

Consider the following scenario:
1. A new image is published by the main pipeline
2. The sync job runs and syncs the new image
3. But line 150, the publish_ssm step fails for some reason, say a random network failure.
4. Next time the sync job runs, it will check if the image sha's match. They will, so it will not publish. SSM will never be updated.

PR #26, opened 2020-03-03
remove echo in setup_test_environment.sh

PettitWesley:


integ/resources/setup_test_environment.sh
```diff
@@ -3,9 +3,9 @@
 # Using CloudFormation describe-stacks extracts the output values for kinesis stream and s3 bucket name, and sets them as environment variables
 stackOutputs=$(aws cloudformation describe-stacks --stack-name integ-test-fluent-bit --output text --query 'Stacks[0].Outputs[*].OutputValue')
 read -r -a outputArray <<< "$stackOutputs"
-echo "Kinesis Stream: ${outputArray[0]}"
-echo "S3 Bucket Name: ${outputArray[1]}"
-echo "Firehose Stream: ${outputArray[2]}"
+echo "Firehose Stream: ${outputArray[0]}"
+echo "Kinesis Stream: ${outputArray[1]}"
+echo "S3 Bucket Name: ${outputArray[2]}"
```
hossain-rayhan: Also, we can remove them if we want. The only purpose was to make sure we got the right data.

PR #30, opened 2020-03-25
Read image version from DockerHub instead of GitHub source

hossain-rayhan:
*Description of changes:*
For the verification process of our sync task, we were reading the image version from AWS_FOR_FLUENT_BIT_VERSION file of our gitHub source. It was causing the sync task to fail. With this change, we will read our latest image version from DockerHub which will solve the above mentioned problem.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -124,7 +124,7 @@ check_parameter() {
 }
 
 sync_latest_image() {
-	LATEST_VERSION= 'curl -s -S "https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/" | jq '."results"[1]["name"]''
+	LATEST_VERSION= 'curl -s -S https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/ | jq '."results"[1]["name"]''
 	echo $LATEST_VERSION
 	region=${1}
 	account_id=${2}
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -18,6 +18,9 @@ scripts=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 cd "${scripts}"
 
 IMAGE_SHA_MATCHED="FALSE"
+AWS_FOR_FLUENT_BIT_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION)
+AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/' | jq '."results"[1]["name"]' | tr -d '"')
```
PettitWesley: I don't think this is safe enough to be used as part of image publishing pipeline. I ran it myself, and it seems that the latest version is always returned. 

However... that only happens because the returned list is sorted with the "latest" tag first and then all of the version tags in order.

I don't think there's any guarantee from the DockerHub API that they will always be sorted that way. If they change their implementation, it will break this script and we will publish the wrong version.

That sort of risk is not acceptable; with a production publishing system like this that is used by tons of customers, we must be able to trust that it will work as expected.

PR #38, opened 2020-05-13
Add logrotate and cron

lorenzomicheli:
*Issue #, if available:*

*Description of changes:*
I have added the following packages:
- logrotate - A log rotation utility
- cronie - A Cron Daemon
- crontabs -  Root crontab files used to schedule the execution of programs 

In the /entrypoint.sh script I explicitly start  crond

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,7 +77,9 @@ RUN yum upgrade -y \
           pkgconfig \
           systemd-devel \
           zlib-devel \
-          nc
+          nc \
+          logrotate \
+          crontabs
 
 COPY --from=builder /fluent-bit /fluent-bit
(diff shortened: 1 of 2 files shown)


PettitWesley:
@lorenzomicheli Why start crond in the entrypoint?

lorenzomicheli:
Otherwise It does not start and if I run it as a service I get the following error:
`bash-4.2# service crond start
Redirecting to /bin/systemctl start crond.service
Failed to get D-Bus connection: Operation not permitted`

PettitWesley:
As discussed, we decided that (at the moment) we don't see sufficient reason to add this to the image. We recommend customers add logrotate to app images instead.

PR #46, opened 2020-06-19
Change base branch from 'master' to 'mainline' in all plugins

PettitWesley:


PettitWesley:
Good point, I'll switch this repo back to master until we update the pipeline

PR #60, opened 2020-07-28
Stop releasing from mainline

hencrice:
*Issue #, if available:*
N/A

*Description of changes:*
We used to release directly from the `mainline` branch of individual plugins. This PR changed it so that the plugins are built from a specific version tag. I've successfully tested the Docker build once with `docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .` before starting to run into all the `rsc.io` issues this morning.

### DIFF
--- a/Dockerfile.plugins
+++ b/Dockerfile.plugins
@@ -12,7 +12,7 @@ ARG CLOUDWATCH_PLUGIN_TAG=v1.3.2
 
 RUN git clone $KINESIS_PLUGIN_CLONE_URL /kinesis-streams
 WORKDIR /kinesis-streams
-RUN git fetch --all --tags && git checkout tags/$KINESIS_PLUGIN_TAG -b $KINESIS_PLUGIN_TAG && git rev-parse HEAD
+RUN git fetch --all --tags && git checkout tags/$KINESIS_PLUGIN_TAG -b $KINESIS_PLUGIN_TAG && git describe --tags
 RUN go mod download
 RUN make release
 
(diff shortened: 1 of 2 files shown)


Dockerfile.plugins
```diff
@@ -1,32 +1,33 @@
...
 ENV GO111MODULE on
 
-# Kinesis Streams
 ARG KINESIS_PLUGIN_CLONE_URL=https://github.com/aws/amazon-kinesis-streams-for-fluent-bit.git
-ARG KINESIS_PLUGIN_BRANCH=mainline
+ARG KINESIS_PLUGIN_TAG=v1.5.1
+ARG FIREHOSE_PLUGIN_CLONE_URL=https://github.com/aws/amazon-kinesis-firehose-for-fluent-bit.git
+ARG FIREHOSE_PLUGIN_TAG=v1.4.1
+ARG CLOUDWATCH_PLUGIN_CLONE_URL=https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit.git
+ARG CLOUDWATCH_PLUGIN_TAG=v1.3.2
```
hossain-rayhan, line 9: I prefer them (tag lines) moving to a different place; or read from the plugin repos.

hossain-rayhan: Otherwise, we can easily miss or mess up.

sonofachamp:
Why are we running ` && git rev-parse HEAD` throughout? If it's just to print the SHA I don't know how useful that is? I'd prefer to see the output of `git describe --tags` after checking out a specific tag.

Dockerfile.plugins
hencrice: Could you elaborate? I don't understand what you meant.

hossain-rayhan: Rather than putting hard coded plugin versions, we can read the latest release tag from original plugin repositories.

hossain-rayhan: If we want to go with this, please add a section in release process doc to update these tags.

Issue #44, opened 2020-06-13
arm64 build needed


otterley:
Now that ECR supports multi-arch images, and EC2 instance types using Graviton2 processors have reached GA, it's time for us to build dual x86-64/arm64 images using multi-arch builds.

JeffUnderhill:
The upstream Fluent Bit project have added Arm64v8 packages for most Linux distributions as a supported platform - [See](https://docs.fluentbit.io/manual/installation/supported-platforms). Are there any other component pieces needed for Arm64 support at this time?

otterley:
This GitHub project produces the Docker image that includes not only the upstream software, but also the AWS-specific plugins it uses (CloudWatch Logs, Kinesis Streams, etc.). Those plugins have to be compiled for Arm64v8, too.

lubingfeng:
This sounds like a feature request rather than a defect to me.

autarchprinceps:
This is particularly important on EKS. This runs as a daemonset to export container logs to CloudWatch Logs, or other targets. If every node runs an instance, but it's not supported on ARM, that means ARM instances aren't properly supported as EKS nodes, not even in a mixed setup with x86 nodes for other non supported software.

lubingfeng:
AWS fluent bit ARM 64 support is coming soon.

zhonghui12:
Hi @otterley @JeffUnderhill @autarchprinceps , ARM64 is included in our recent release. Please check here: https://github.com/aws/aws-for-fluent-bit/blob/mainline/CHANGELOG.md.

Thanks. This issue will be closed.

PR #71, opened 2020-09-15
Support for ARM64 build

MeghnaPrabhu:
Support for ARM64 build
https://github.com/aws/aws-for-fluent-bit/issues/44

*Description of changes:*

- PR does not include the changes for sync task. I will publish a separate PR for that. 
- Pipeline stages have been modified to allow for parallel multi-arch image builds (x86-64 and ARM64).

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -14,9 +14,24 @@ phases:
       # List the docker images 
       - docker images
 
-      # Push the image to ECR in the same account and same region the pipeline is hosted.
-      - docker tag amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit-test:latest
-      - ecs-cli push amazon/aws-for-fluent-bit-test:latest
+      # Push the image to ECR with corresponding architecture as the tag.
+      - aws ecr get-login-password --region ${AWS_REGION}| docker login --username AWS --password-stdin ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com
+      - aws ecr create-repository --repository-name amazon/aws-for-fluent-bit-test --image-scanning-configuration scanOnPush=true --region ${AWS_REGION}  || true
+      - architecture=$(docker inspect --format='{{.Architecture}}'  amazon/aws-for-fluent-bit)
+      - docker tag amazon/aws-for-fluent-bit:latest ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$architecture
+      - docker push ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$architecture
+
+       # Create manifest list
+      - export DOCKER_CLI_EXPERIMENTAL=enabled
+      - docker manifest create ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:arm64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:amd64 || true
+      - docker manifest annotate --arch arm64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:arm64 || true
+      - docker manifest annotate --arch amd64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:amd64 || true
+      
+      # Sanity check for the debug log
+      - docker manifest inspect ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest || true
+
+      # Push manifest list
+      - docker manifest push ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest || true 
 artifacts:
   files:
     - '**/*'
--- a/integ/resources/create_test_resources.sh
+++ b/integ/resources/create_test_resources.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
 # Deploys the CloudFormation template to create the stack and necessary resources- kinesis data stream, s3 bucket, and kinesis firehose delivery stream
-# Resource (stream, s3, delivery stream) names will start with the stack name "integ-test-fluent-bit"
-aws cloudformation deploy --template-file ./integ/resources/cfn-kinesis-s3-firehose.yml --stack-name integ-test-fluent-bit --region us-west-2 --capabilities CAPABILITY_NAMED_IAM
+# Resource (stream, s3, delivery stream) names will start with the stack name followed by the corresponding architecture. "integ-test-fluent-bit-architecture"
+ARCHITECTURE=$(uname -m)
+if [ ARCHITECTURE=="x86_64" ] 
+then
+    ARCHITECTURE="x86-64"
+fi
+aws cloudformation deploy --template-file ./integ/resources/cfn-kinesis-s3-firehose.yml --stack-name integ-test-fluent-bit-${ARCHITECTURE} --region us-west-2 --capabilities CAPABILITY_NAMED_IAM
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -73,8 +73,11 @@ gamma_account_id="626332813196"
 
 DOCKER_HUB_SECRET="com.amazonaws.dockerhub.aws-for-fluent-bit.credentials"
 
+ARCHITECTURES=("amd64" "arm64")
+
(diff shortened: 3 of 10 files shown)


scripts/publish.sh
```diff
@@ -43,38 +43,41 @@ eu-west-3
...
-classic_regions_account_id="906394416424"
+classic_regions_account_id="957584016365"
 
 cn_regions="
 cn-north-1
 cn-northwest-1
 "
 
-cn_regions_account_id="128054284489"
+cn_regions_account_id="957584016365"
```
PettitWesley, line 54: Guessing this is left over from testing- account IDs should not change

MeghnaPrabhu: Yes, I will switch these from my personal ID as the last commit before merge.

buildspec.yml
```diff
@@ -14,9 +14,28 @@ phases:
       # List the docker images 
       - docker images
 
-      # Push the image to ECR in the same account and same region the pipeline is hosted.
-      - docker tag amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit-test:latest
-      - ecs-cli push amazon/aws-for-fluent-bit-test:latest
+      # Push the image to ECR with corresponding architecture as the tag.
+      - aws ecr get-login-password --region ${AWS_REGION}| docker login --username AWS --password-stdin ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com
+      - aws ecr create-repository --repository-name amazon/aws-for-fluent-bit-test --image-scanning-configuration scanOnPush=true --region ${AWS_REGION}  || true
+      - architecture=$(docker inspect --format='{{.Architecture}}'  amazon/aws-for-fluent-bit)
```
hossain-rayhan, line 20: I am not super clear here. For the first time, we only have AMD version image in dockerhub. How will we get the proper ARM architecture using this command. Excuse my ignorance with docker inspect.

MeghnaPrabhu: So in the build stage, we have two instances AMD and ARM running in parallel. Both instances will build the image separately, in parallel. Before these images are pushed to ECR, I tag each image with it's underlying architecture using  `architecture=$(docker inspect --format='{{.Architecture}}'  amazon/aws-for-fluent-bit)`. So as you can see, order does not matter here. 

However, when we create the manifest list after this step, we need to wait for both the images to be available in ECR which is why after each command, I have a `|| True`. This ensures build does not fail.

buildspec.yml
```diff
@@ -14,9 +14,28 @@ phases:
...
+      - docker manifest annotate --arch arm64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest \
+          ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:arm64 || true
+      - docker manifest annotate --arch amd64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest \
+          ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:amd64 || true
+      
+      # Sanity check for the debug log
+      - docker manifest inspect ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest || true
+
+      # Push manifest list
+      - docker manifest push ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:latest || true
```
hossain-rayhan, line 38: Why do we need to do a `OR` with true ( || true) every time? Shouldn't it return an error code if it fails?

MeghnaPrabhu: Like I mentioned above, since both instances are running in parallel, only the one that executes last will be able to create the manifest list successfully. We need both the images to be pushed to ECR before the manifest list is created. At the same time, build should not fail if it is waiting for the image of the other architecture to be available, this is why we have the `|| True`

scripts/publish.sh
```diff
@@ -259,7 +308,7 @@ match_two_sha() {
 
 if [ "${1}" = "publish" ]; then
 	if [ "${2}" = "dockerhub" ]; then
-		publish_to_docker_hub amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_VERSION}
+		publish_to_docker_hub amazon/aws-for-fluent-bit meghnaprabhu/aws-for-fluent-bit
```
hossain-rayhan, line 311: Need to replace with `amazon`.

scripts/publish.sh
```diff
@@ -417,7 +466,7 @@ fi
 # Following scripts will be called only from the CI/CD pipeline
 if [ "${1}" = "cicd-publish" ]; then
 	if [ "${2}" = "dockerhub" ]; then
-		publish_to_docker_hub amazon/aws-for-fluent-bit:latest amazon/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_VERSION}
+		publish_to_docker_hub amazon/aws-for-fluent-bit meghnaprabhu/aws-for-fluent-bit
```
hossain-rayhan, line 469: Same here.

scripts/publish.sh
MeghnaPrabhu: Yes, I will replace that as the last step along with the account ID

hossain-rayhan: Also, I can see both of the values are same. If we are not using the 2nd parameter (fluent-bit-version), we can remove it.

hossain-rayhan:
Looks good to me. Once minor feedbacks are addressed, we are good to go. 
One thing, I didn't see any changes for the verification. Is there no way to make sure customers are getting the proper ARM64 image?

MeghnaPrabhu:
For verification did you mean verify if docker pull from a manifest list actually gives you ARM image or did you mean verification of the SHA between tags AWS_FOR_FLUENT_BIT_VERSION and latest?

hossain-rayhan:
The first one. We are not verifying from an ARM environment, right?

MeghnaPrabhu:
That is guaranteed by docker, if you issue a docker pull on a manifest file, docker will return the image compatible with the underlying architecture.

hossain-rayhan (approved):
Looks good to me.

buildspec.yml
PettitWesley: That's fine, but then we need the verify step to check that the manifest list was published properly. Otherwise its possible for both instances to fail without us knowing, right?

MeghnaPrabhu: That is a good point, I thought about it.  However, if the manifest list does not get created properly, it will fail immediately in the integ-test stage. This is because we start each stage by pulling the image from the manifest list that was created during the build. 

Adding a verification step in the publish to ecr stage will not be very useful because by then it is established that we have the manifest file working correctly.

MeghnaPrabhu: Additionally, this case of `||true` only occurs in the build stage because we are running both instances in parallel. Other stages like publish to ecr are all running on a single instance.  Hence, we can be sure that the manifest list created in the publish to ecr stage has both the images as we will be pushing them individually to ECR before creating the manifest.

PettitWesley: Makes sense 👍

integ/resources/create_test_resources.sh
```diff
@@ -1,5 +1,10 @@
...
 # Deploys the CloudFormation template to create the stack and necessary resources- kinesis data stream, s3 bucket, and kinesis firehose delivery stream
-# Resource (stream, s3, delivery stream) names will start with the stack name "integ-test-fluent-bit"
-aws cloudformation deploy --template-file ./integ/resources/cfn-kinesis-s3-firehose.yml --stack-name integ-test-fluent-bit --region us-west-2 --capabilities CAPABILITY_NAMED_IAM
+# Resource (stream, s3, delivery stream) names will start with the stack name followed by the corresponding architecture. "integ-test-fluent-bit-architecture"
+ARCHITECTURE=$(uname -m)
+if [ ARCHITECTURE=="x86_64" ] 
+then
+    ARCHITECTURE="x86-64"
+fi
+aws cloudformation deploy --template-file ./integ/resources/cfn-kinesis-s3-firehose.yml --stack-name integ-test-fluent-bit-${ARCHITECTURE} --region us-west-2 --capabilities CAPABILITY_NAMED_IAM
```
PettitWesley, line 10: Suggestion, replace with if with:

```
ARCHITECTURE=$(uname -m | tr '_' '-')
```

scripts/publish.sh
```diff
@@ -173,11 +186,41 @@ verify_ssm() {
...
+	# TODO: Add a way to automatically generate arch images in manifest 
+	docker manifest create ${1}:${tag} \
+		${1}:arm64-${AWS_FOR_FLUENT_BIT_VERSION} \
+		${1}:amd64-${AWS_FOR_FLUENT_BIT_VERSION} 
+
+	for arch in "${ARCHITECTURES[@]}"
+	do
+    docker manifest annotate --arch "$arch" \
+		${1}:${tag} \
+		${1}:"$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
```
PettitWesley, line 203: weird indent on one line

scripts/publish.sh
```diff
@@ -189,8 +232,17 @@ make_repo_public() {
...
+	echo $region
+	echo $account_id
+
+	aws ecr get-login-password --region ${region}| docker login --username AWS --password-stdin ${account_id}.dkr.ecr.${region}.amazonaws.com
+	aws ecr create-repository --repository-name aws-for-fluent-bit --image-scanning-configuration scanOnPush=true --region ${region}  || true
+	
+	push_image_ecr ${account_id} ${region}	
+
+    create_manifest_list ${account_id}.dkr.ecr.${region}.amazonaws.com/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_VERSION}
+	create_manifest_list ${account_id}.dkr.ecr.${region}.amazonaws.com/aws-for-fluent-bit "latest"
```
PettitWesley, line 244: weird indent on one line

PettitWesley (approved):

PR #75, opened 2020-09-18
Add integration tests for out_s3

PettitWesley:
These can't be merged until out_s3 is released.

### DIFF
--- a/integ/s3-logger/logscript.sh
+++ b/integ/s3-logger/logscript.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Writes 7717 unique log lines
+# intermixed with 1 KB lines of random data
+# these logs are written over the course of a little over 1 minute
+# because we want to test S3's local buffering
+# Why 7717? Its a prime number. And I like that.
+# Finally we sleep for 90s- ensuring the upload timeout hits for the last chunk
+# then exit
+
+for i in {0..7716}
+do
+    echo $i
+    openssl rand -base64 1000 | tr '\n' '-' && echo ""
+    sleep 0.0001
+done
+
+sleep 90
+
+exit 0
--- a/integ/s3-logger/logscript.sh
+++ b/integ/s3-logger/logscript.sh
@@ -15,6 +15,6 @@ do
     sleep 0.0001
 done
 
-sleep 90
+sleep 200
 
 exit 0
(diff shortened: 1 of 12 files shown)


integ/s3/validate-and-clean.go
```diff
@@ -121,6 +131,9 @@ func validate(s3Client *s3.S3, response *s3.ListObjectsV2Output, bucket string,
 			if d == "" {
 				continue
 			}
+			if len(d) > 500 {
+				continue
```
hossain-rayhan, line 135: Maybe I am missing something. Why do we need this condition and the upper limit is 500?

integ/s3-logger/logscript.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Writes 7717 unique log lines
+# intermixed with 1 KB lines of random data
+# these logs are written over the course of a little over 1 minute
+# because we want to test S3's local buffering
+# Why 7717? Its a prime number. And I like that.
```
hossain-rayhan, line 7: lol- I was searching for it.

hossain-rayhan:
Overall looks good to me. Just one point- do we need to do anything specific for parallel testing (AMD64 and ARM64). I remember we had an issue with CloudWatch test. I don't think so because its using the same validation method as Kinesis and Firehose plugin.

integ/s3/validate-and-clean.go
PettitWesley: Look at the new S3 logger, it spits out log lines that are just numbers, and also large lines of 1KB random data (because S3 can only send data in chunks of a few MB). I wanted to make it so that the code didn't waste time trying to process those lines with random data.

500 is an arbitrary number between the size of each type of log event. The integer lines that we count will be smaller and the random data lines will be larger.

PettitWesley:
> Just one point- do we need to do anything specific for parallel testing (AMD64 and ARM64).

Do the arm changes run the integ tests in parallel? I thought it was just build. @meghnapr 

If the tests will be run in parallel I will have to change the S3 key prefix in the test so that its different for each achitecture.

MeghnaPrabhu:
Yes, we are running integ tests in parallel so the S3 key prefix needs to be appended with the architecture.

PettitWesley:
Ive add the architecture in the S3 key so that this will work with arm64 support

hossain-rayhan (approved):

PR #76, opened 2020-09-23
Support for ARM64 on sync task 

MeghnaPrabhu:
Support for ARM64 build https://github.com/aws/aws-for-fluent-bit/issues/44

Added support to ensure that multi arch images are available in special ECR regions. 

Description of changes:

- Changes for sync task are only made in publish.sh https://github.com/aws/aws-for-fluent-bit/commit/a5b3d1150a524865dc01aab5670ec214034c6c00
- PR includes previous commit https://github.com/aws/aws-for-fluent-bit/pull/73 for ease of merge.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #73, opened 2020-09-17
Support for ARM64 build

MeghnaPrabhu:
Support for ARM64 build
https://github.com/aws/aws-for-fluent-bit/issues/44

*Description of changes:*

- PR does not include the changes for sync task. I will publish a separate PR for that.
- Pipeline stages have been modified to allow for parallel multi-arch image builds (x86-64 and ARM64).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #77, opened 2020-09-25
Ecr image scanning for CI/CD

MeghnaPrabhu:
Enabling ECR Image scanning on the updated pipeline.
Every time an image is built and pushed to ECR we will scan and report vulnerabilities. If there are vulnerabilities, the pipeline will immediately fail.

*Description of changes:*

- added ecr image scan verification on the updated pipeline for multi-arch build. 
   Hence this PR has changes from https://github.com/aws/aws-for-fluent-bit/pull/73 and https://github.com/aws/aws-for-fluent-bit/pull/76
- pipeline will fail if scan returns vulnerabilities

For ease of review, here are the changes specific to image scan

- https://github.com/MeghnaPrabhu/aws-for-fluent-bit/blob/ecr-image-scan/scripts/publish.sh#L280
- https://github.com/MeghnaPrabhu/aws-for-fluent-bit/blob/ecr-image-scan/scripts/publish.sh#L581
- https://github.com/MeghnaPrabhu/aws-for-fluent-bit/blob/ecr-image-scan/scripts/publish.sh#L179
- https://github.com/MeghnaPrabhu/aws-for-fluent-bit/blob/ecr-image-scan/scripts/publish.sh#L247
- https://github.com/MeghnaPrabhu/aws-for-fluent-bit/blob/ecr-image-scan/buildspec.yml#L23

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -176,7 +176,7 @@ sync_latest_image() {
 		if [ "$IMAGE_SHA_MATCHED" = "FALSE" ]; then
 			aws ecr create-repository --repository-name aws-for-fluent-bit --image-scanning-configuration scanOnPush=true --region ${region}  || true
 			push_image_ecr amazon/aws-for-fluent-bit:${arch}-${AWS_FOR_FLUENT_BIT_VERSION} \
-				${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:${arch}-${AWS_FOR_FLUENT_BIT_VERSION}
+				${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:${arch}-${AWS_FOR_FLUENT_BIT_VERSION} 
 		fi
 	done
 
@@ -272,6 +272,24 @@ verify_ecr() {
 	fi
 
 	verify_sha $sha1 $sha2
+}
+
+verify_ecr_image_scan() {
+	region=${1}
+	repo_uri=${2}
+	tag=${3}
+	
+	imageTag=$(aws ecr list-images  --repository-name ${repo_uri} --region ${region} | jq -r '.imageIds[].imageTag' | grep -c ${tag} || echo "0")
+	if [ "$imageTag" = '1' ]; then
+		aws ecr start-image-scan --repository-name ${repo_uri} --image-id imageTag=${tag} --region ${region}	
+		aws ecr wait image-scan-complete --repository-name ${repo_uri} --region ${region} --image-id imageTag=${tag}
+		highVulnerabilityCount=$(aws ecr describe-image-scan-findings --repository-name ${repo_uri} --region ${region} --image-id imageTag=${tag} | jq '.imageScanFindings.findingSeverityCounts.HIGH')
+		criticalVulnerabilityCount=$(aws ecr describe-image-scan-findings --repository-name ${repo_uri} --region ${region} --image-id imageTag=${tag} | jq '.imageScanFindings.findingSeverityCounts.CRITICAL')
+		if [ "$highVulnerabilityCount" != null ] || [ "$criticalVulnerabilityCount" != null ]; then
+			echo "Uploaded image ${tag} has ${vulnerabilityCount} vulnerabilities."
+			exit 1
+		fi	
+	fi
 }
 
 verify_dockerhub() {
@@ -552,4 +570,8 @@ if [ "${1}" = "cicd-verify-ssm" ]; then
 			verify_ssm ${region}
 		done
 	fi
+fi
+
+if [ "${1}" = "cicd-verify-ecr-image-scan" ]; then
+	verify_ecr_image_scan ${2} ${3} ${4}	
 fi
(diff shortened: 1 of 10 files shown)


scripts/publish.sh
```diff
@@ -271,6 +277,23 @@ verify_ecr() {
...
+	tag=${3}
+	
+	imageTag=$(aws ecr list-images  --repository-name ${repo_uri} --region ${region} | jq -r '.imageIds[].imageTag' | grep -c ${tag} || echo "0")
+	if [ "$imageTag" = '1' ]; then
+		aws ecr start-image-scan --repository-name ${repo_uri} --image-id imageTag=${tag} --region ${region}	
+		aws ecr wait image-scan-complete --repository-name ${repo_uri} --region ${region} --image-id imageTag=${tag}
+		vulnerabilityCount=$(aws ecr describe-image-scan-findings --repository-name ${repo_uri} --region ${region} --image-id imageTag=${tag} | jq '.imageScanFindings.findings | length')
+		if [ "$vulnerabilityCount" != '0' ]; then
+			echo "Uploaded image ${tag} has ${vulnerabilityCount} vulnerabilities."
+			exit 1
```
PettitWesley, line 292: Looking at the docs: https://docs.aws.amazon.com/cli/latest/reference/ecr/describe-image-scan-findings.html

You get this:
```
"findingSeverityCounts": {
          "MEDIUM": 1
      }
```

So we can do per-severity counts, and only block certain severities. (Not sure what we should restrict it to- may be high and critical?)

MeghnaPrabhu: Agreed, as per this https://nvd.nist.gov/vuln-metrics/cvss, I think blocking high and critical should be good.

scripts/publish.sh
```diff
@@ -239,6 +242,9 @@ publish_ecr() {
 	do
 		push_image_ecr ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:"$arch" \
 			${account_id}.dkr.ecr.${region}.amazonaws.com/aws-for-fluent-bit:"$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
+		
+		# Added verification for sanity check even though we are pushing the same image that we scanned in build stage.
+		verify_ecr_image_scan ${region} aws-for-fluent-bit "$arch"-${AWS_FOR_FLUENT_BIT_VERSION} 
 	done
```
PettitWesley: Let's not do this. I think we should only scan in one part of the build, no where else. 

Otherwise I think there's a small chance we could get caught in a weird situation. Imagine in build the image scan passes, but then a new vulnerability is added to the database as the pipeline moves on, and then suddenly we've published to some regions but not others.

scripts/publish.sh
```diff
@@ -174,6 +174,9 @@ sync_latest_image() {
 			aws ecr create-repository --repository-name aws-for-fluent-bit --image-scanning-configuration scanOnPush=true --region ${region}  || true
 			push_image_ecr amazon/aws-for-fluent-bit:${arch}-${AWS_FOR_FLUENT_BIT_VERSION} \
 				${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:${arch}-${AWS_FOR_FLUENT_BIT_VERSION}
+
+			# Added verification for sanity check.
+			verify_ecr_image_scan ${region} aws-for-fluent-bit "$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
```
PettitWesley, line 179: See comment below

scripts/publish.sh
MeghnaPrabhu: If I understand correctly, we build the image and scan it to make sure there are no vulnerabilities. Then we push this same image to various regions and make the images public. So where would the new vulnerability be added in the pipeline, I guess I am missing something. Can you explain? Also is it possible that the same image in one region could be good but in another region bad? 

As I mentioned in the comments, it is not necessary for us to do this check as we have already scanned the image but I felt it was a good sanity check just before we make the image public. However as you mentioned, if a vulnerability creeps in, I think it is a good idea to have this in place so that we catch it early. 

Additionally, our goal is to perform image scan more frequently (1 scan per day). At this point, we will not build a new image, but scan the existing ones. So the issue you mentioned where having some regions with bad images could occur. How do we deal with that?

PettitWesley: > So where would the new vulnerability be added in the pipeline, I guess I am missing something. 

I think you're thinking of the vulnerability as something which only exists in the image. Which is technically true. But the vulnerability is only discovered if it exists in both the image and in the database of vulnerabilities, which is always being updated. 

So its possible you could have an image with a vulnerability that is unknown at build time, but that becomes "known" in the hour or so it takes for the pipeline and sync tasks to publish everywhere. Which then means that we only publish to some regions, which makes the rollback and messaging to customers complicated.

MeghnaPrabhu: I see now how it can be a problem. Thanks for this.

PettitWesley (approved):

PR #79, opened 2020-10-12
Force the use of STS regional endpoints

PettitWesley:
It turns out that the AWS CLI only uses regional STS endpoints if you set a env var: https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-sts_regional_endpoints.html

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -20,9 +20,12 @@ cd "${scripts}"
 IMAGE_SHA_MATCHED="FALSE"
 AWS_FOR_FLUENT_BIT_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION)
 
-docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/' | jq -r '.results[].name' | sort -r) 
+docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/' | jq -r '.results[].name' | sort -r)
 tag_array=(`echo ${docker_hub_image_tags}`)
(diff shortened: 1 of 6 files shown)


scripts/publish.sh
```diff
@@ -20,10 +20,13 @@ cd "${scripts}"
 IMAGE_SHA_MATCHED="FALSE"
 AWS_FOR_FLUENT_BIT_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION)
 
-docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/' | jq -r '.results[].name' | sort -r) 
+docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/' | jq -r '.results[].name' | sort -r)
 tag_array=(`echo ${docker_hub_image_tags}`)
 AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=${tag_array[1]}
 
+# Enforce STS regional endpoints
+AWS_STS_REGIONAL_ENDPOINTS=regional
```
hossain-rayhan, line 28: Where are we using this value?

PettitWesley: AWS CLI uses it, as shown in the doc link I put in the description: https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-sts_regional_endpoints.html

hossain-rayhan: May be I am missing something. To me, it seems like a normal class variable, not an environment variable. Where it's being utilized in this file?

hossain-rayhan: It's a `.sh` file and it makes sense I guess.

hossain-rayhan (approved):

Issue #90, opened 2020-10-26
Upgrade to Fluent-bit 1.6.2 - Fix ElasticSearch


allamand:
See : https://fluentbit.io/announcements/v1.6.2/
I'm interested on the Fix regarding ElasticSearch

PettitWesley:
We will do a release some time this week

allamand:
Thanks @PettitWesley

zhonghui12:
Hi @allamand, we have upgraded to Fluent-bit 1.6.3 for recent release. Please check more details here: https://github.com/aws/aws-for-fluent-bit/blob/mainline/CHANGELOG.md.

Will close this issue. Thanks.

Issue #81, opened 2020-10-14
Some logs are not created in Cloudwatch with version 2.8.0


caldo-pys:
Hi,
I have been using aws-for-fluent-bit:latest in production since few months with no issue, with redirection of logs to cloudwatch.
Since few days, it seems like some logs are not written anymore in Cloudwatch.
I went back to aws-for-fluent-bit:2.7.0 and it works perfectly.
I tried with aws-for-fluent-bit:2.8.0 and I am experiencing the issue

Here is the log of 2.7.0 version:

>  | 2020-10-14T17:18:25.438+04:00 | AWS for Fluent Bit Container Image Version 2.7.0
>   | 2020-10-14T17:18:25.483+04:00 | [1mFluent Bit v1.5.6[0m
>   | 2020-10-14T17:18:25.483+04:00 | * [1m[93mCopyright (C) 2019-2020 The Fluent Bit Authors[0m
>   | 2020-10-14T17:18:25.483+04:00 | * [1m[93mCopyright (C) 2015-2018 Treasure Data[0m
>   | 2020-10-14T17:18:25.483+04:00 | * Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
>   | 2020-10-14T17:18:25.483+04:00 | * https://fluentbit.io
>   | 2020-10-14T17:18:25.484+04:00 | time="2020-10-14T13:18:25Z" level=info msg="[cloudwatch 0] plugin parameter log_group = <log_group>"
>   | 2020-10-14T17:18:25.484+04:00 | time="2020-10-14T13:18:25Z" level=info msg="[cloudwatch 0] plugin parameter log_stream_prefix = 'echo hello/'"
>   | 2020-10-14T17:18:25.484+04:00 | time="2020-10-14T13:18:25Z" level=info msg="[cloudwatch 0] plugin parameter log_stream_name = ''"
>   | 2020-10-14T17:18:25.485+04:00 | time="2020-10-14T13:18:25Z" level=info msg="[cloudwatch 0] plugin parameter region = 'eu-central-1'"
>   | 2020-10-14T17:18:25.485+04:00 | time="2020-10-14T13:18:25Z" level=info msg="[cloudwatch 0] plugin parameter log_key = 'log'"
>   | 2020-10-14T17:18:25.485+04:00 | time="2020-10-14T13:18:25Z" level=info msg="[cloudwatch 0] plugin parameter role_arn = ''"
[...]

PR #91, opened 2020-10-26
Upgrade to Fluent-bit 1.6.2

allamand:
Signed-off-by: Sébastien Allamand <sallaman@amazon.com>

*Issue #, if available:* #90 

*Description of changes:*

Upgrade to Fluent-bit 1.6.2 which provide several fixes see : https://fluentbit.io/announcements/v1.6.2/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.9.0
+This release includes:
+* Fluent Bit [1.6.2](https://fluentbit.io/announcements/v1.6.2/)
+
 ### 2.8.0
 This release includes:
 * An Amazon Linux 2 Base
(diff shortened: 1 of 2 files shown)


PettitWesley:
So the reason why we haven't updated yet is because there is also this bug: https://github.com/aws/aws-for-fluent-bit/issues/81

Which we are working on fixing and hoping to get out in the next Fluent Bit release (1.6.3) in a few days.

I am very open to feedback on this- do you think we should have created a release with 1.6.2 already, even though it has known bugs (though less bugs than 1.6.2), instead of waiting till we get everything fixed?

allamand:
I think it a matter of time and benefits. If you say that it as already less bug, I woulad say it makes sense to publish new version already. In the meantime if it only matter of few day to wait for the correction this cal also makes sense.

For my side, i'm waiting for other corrections than this one, but there are lots of things i didn't see or measures the impacts, So I'll trust your choices :)

PettitWesley:
Superceded by #92 which will upgrade to 1.6.3

PR #101, opened 2020-11-10
rm deeplink anchor to examples repo

tedder:
*Issue #, if available:*

*Description of changes:*

This anchor doesn't exist or was renamed. I could use an anchor but it's probably best to decouple it so a browser doesn't dump the user at the bottom of the page.


hossain-rayhan:
It's working for me. Which browser/version you are using?

tedder:
Hmm. That's from modern Chrome on MacOS. But maybe it was my behavior that caused it to go elsewhere, when I try in incognito it's fine.

Here's what the anchor tag looks like in the other repo:
```
<h3><a id="user-content-fluent-bit-examples" class="anchor" aria-hidden="true" href="#fluent-bit-examples">
```

[Here's a random gist talking about it](https://gist.github.com/borekb/f83e48479aceaafa43108e021600f7e3); I don't know where authoritative or canonical documentation is.

Issue #123, opened 2020-12-16
Feature Request: Publish image onto ECR Public


furikake:
There are no images published on ECR Public Gallery: https://gallery.ecr.aws/aws-observability/aws-for-fluent-bit 

It would be good to publish it there so that we can simply use `public.ecr.aws/aws-observability/aws-for-fluent-bit:tag` as our container image.

Doing `aws ssm get-parameters-by-path --path /aws/service/aws-for-fluent-bit/` to parse the output is not user friendly at all.

PettitWesley:
@furikake agreed- we created that repo since are working on this :)

The winter holidays and post-re:invent haze will probably slow things down a bit, but you'll see images there soon enough.

zhonghui12:
Hello @furikake, the feature will be available in next release. Thanks for your patience.

furikake:
Thanks @zhonghui12

zhonghui12:
Hello @furikake, our images have been pushed to Amazon ECR Public Gallery in our recent release: https://github.com/aws/aws-for-fluent-bit/releases/tag/v2.10.1. You can check the public repo to see more details: https://gallery.ecr.aws/aws-observability/aws-for-fluent-bit.

The issue will be closed. Thanks.

PR #124, opened 2020-12-16
Publish image to Public ECR

zhonghui12:
*Issue #123 , if available:*

*Description of changes:*
Publish image onto ECR Public

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -118,6 +118,20 @@ publish_to_docker_hub() {
 
 }
 
+publish_to_public_ecr() {
+	aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
+
+	for arch in "${ARCHITECTURES[@]}"
+	do
+		docker tag ${1}:"$arch" public.ecr.aws/aws-observability/aws-for-fluent-bit:"$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
+		docker push public.ecr.aws/aws-observability/aws-for-fluent-bit:"$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
+	done
+
+	create_manifest_list public.ecr.aws/aws-observability/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_VERSION} ${AWS_FOR_FLUENT_BIT_VERSION}
+	aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
+	create_manifest_list public.ecr.aws/aws-observability/aws-for-fluent-bit "latest" ${AWS_FOR_FLUENT_BIT_VERSION}
+}
+
 publish_ssm() {
 	aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/${3} --overwrite \
 		--description 'Regional Amazon ECR Image URI for the latest AWS for Fluent Bit Docker Image' \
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -118,6 +118,20 @@ publish_to_docker_hub() {
 
 }
 
+publish_to_public_ecr() {
+	aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
+	aws ecr-public create-repository --repository-name aws-for-fluent-bit --catalog-data file://awsforfluentbit.json --region us-east-1  || true
```
PettitWesley, line 123: I don't think this is needed, the repo should already exist

zhonghui12: Yeah.. Make sense. Will remove this line.

scripts/publish.sh
```diff
@@ -46,7 +46,7 @@ eu-west-3
 eu-north-1
 "
 
-classic_regions_account_id="906394416424"
+classic_regions_account_id="075490442118"
```
hossain-rayhan, line 49: Did we create a new account?

zhonghui12: No, this is my account.. for testing purpose. Will modify this part after the test.

scripts/publish.sh
```diff
@@ -118,6 +118,20 @@ publish_to_docker_hub() {
 
 }
 
+publish_to_public_ecr() {
+	aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
```
hossain-rayhan, line 122: I am not sure why its default to `us-east-1`? Is it for public ECE?

zhonghui12: Yes, for public ECR, we always set the region to `us-east-1`

scripts/publish.sh
hossain-rayhan: Ahh kind of confused? Can't we test before merging the code into mainline?

scripts/publish.sh
PettitWesley, line 122: for testing, use a different public repo. Create one in your own account and make the code publish to it.

scripts/publish.sh
hossain-rayhan: Ahh kind of confused. Can't we test before merging the code into mainline?

zhonghui12: Yes, we can. But for this PR, I didn't mean to merge it now. This is not the final version for it and this PR is just a way for members to leave comments. Sorry for the confusion..

scripts/publish.sh
hossain-rayhan: Maybe I don't have enough insights here. For public ECR is it necessary to hit a regional endpoint for getting login-password?

zhonghui12: Yes we should: https://docs.aws.amazon.com/AmazonECR/latest/public/getting-started-cli.html. We still need to authenticate to the public registry but specify `--region us-east-1`.

scripts/publish.sh
```diff
@@ -118,6 +118,20 @@ publish_to_docker_hub() {
...
+
+	for arch in "${ARCHITECTURES[@]}"
+	do
+		docker tag ${1}:"$arch" public.ecr.aws/aws-observability/aws-for-fluent-bit:"$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
+		docker push public.ecr.aws/aws-observability/aws-for-fluent-bit:"$arch"-${AWS_FOR_FLUENT_BIT_VERSION}
+	done
+
+	create_manifest_list public.ecr.aws/aws-observability/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_VERSION} ${AWS_FOR_FLUENT_BIT_VERSION}
+	aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/aws-observability
```
hossain-rayhan, line 131: Do we need this again? We got the login credentials in line 122.

zhonghui12: yes we need it. For some reason, we need to got it every time before we want to `create manifest list`. If we don't add this line, we would have an error.

hossain-rayhan: nit: Hmm, interesting. [Here](https://github.com/aws/aws-for-fluent-bit/pull/124/files#diff-c63759f58a022cf694046a961470c6e28c042d8f221646bc5ddbb5fa959dda75R255) they did the same but for regional ECR I guess. Do we have the error log?

zhonghui12: yes we have.
Screen Shot 2020-12-18 at 1 55 43 PM

zhonghui12: You can see here. I didn't log in again in my previous code and it succeeds with `tag: 2.10.0` but failed with `tag:latest`. The error info is `denied: Not Authorized`.

hossain-rayhan:
I think we also need to publish the SSM parameter for the ECR public repo. We don't need any additional pipeline stage but have to update the publish_ssm script.

PettitWesley:
@hossain-rayhan @zhonghui12  I don't think there's any value in publishing an SSM public parameter for the public repo.

The SSM parameters are needed for the private repos since they contain an opaque account ID which is different in each region. For the public repo, the link never changes, so we don't need a parameter.

scripts/publish.sh
```diff
@@ -499,6 +525,8 @@ fi
 if [ "${1}" = "cicd-publish" ]; then
 	if [ "${2}" = "dockerhub" ]; then
 		publish_to_docker_hub amazon/aws-for-fluent-bit
+	elif [ "${2}" = "public-ecr" ]; then
+		publish_to_public_ecr amazon/aws-for-fluent-bit
```
PettitWesley, line 529: nit: is the argument `amazon/aws-for-fluent-bit` to the function actually needed? I think it isn't

PettitWesley (approved):

PettitWesley (approved):
Needs @hossain-rayhan review/approval

hossain-rayhan (approved):

PR #139, opened 2021-02-01
Add a stable tag for the release to mark the latest stable version

zhonghui12:
*Issue #, if available:*

*Description of changes:*


zhonghui12:
I think two things should be noticed here:
1. we can only set the stable version after 2.9.0 because that's the beginning version supports arm64.
2. we may need to consider push the stable image to our docker in classic regions. I didn't implement that part here and I am just thinking about if we need to do that

PettitWesley:
> we may need to consider push the stable image to our docker in classic regions. I didn't implement that part here and I am just thinking about if we need to do that

*Docker Hub not Docker.

I agree. I think in the classic region sync task we need to publish the stable tag to Public ECR and Docker Hub.

buildspec_sync.yml
```diff
@@ -24,8 +24,8 @@ phases:
       - export AWS_SESSION_TOKEN=`echo $CREDS | jq -r .Credentials.SessionToken`
 
       # Verify from the verification account
-      - './scripts/publish.sh cicd-verify ${AWS_REGION}'
-      - './scripts/publish.sh cicd-verify-ssm ${AWS_REGION}'
+      - './scripts/publish.sh cicd-verify ${AWS_REGION} stable'
+      - './scripts/publish.sh cicd-verify-ssm ${AWS_REGION} stable'
 artifacts:
```
PettitWesley, line 29: What does adding this new argument accomplish? The code does not seem to use it as far as I can tell. I don't think its needed? Since the sync tasks always check stable version now.

zhonghui12: because we need to know how to set the third parameter here: https://github.com/zhonghui12/aws-for-fluent-bit/blob/pr-stable-tagging/scripts/publish.sh#L611. The same in method `cicd-verify-ssm`

zhonghui12:
This PR will be closed since it is involved in another one: https://github.com/aws/aws-for-fluent-bit/pull/141.

PR #152, opened 2021-02-24
Remove /var/cache/yum to reduce the size of the container image

### LINKED ISSUE #17: Feature request: tiny up docker image
Currently image is around 553Mb which makes difficult to use as for example pod sidecar. Can we make it a reasonable or size without all unnecessary software installed. It can be multistage build for example

kzys:
closes #17 

We don't assume customers update RPM packages inside the container.
Removing the cache directory makes the image ~200MB smaller.

Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

*Issue #, if available:*

NA

*Description of changes:*

Removes /var/cache/yum to make the container image smaller.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ RUN yum upgrade -y \
           pkgconfig \
           systemd-devel \
           zlib-devel \
-          nc
+          nc && rm -fr /var/cache/yum
 
 COPY --from=builder /fluent-bit /fluent-bit
 COPY --from=aws-fluent-bit-plugins:latest /kinesis-streams/bin/kinesis.so /fluent-bit/kinesis.so


Dockerfile
```diff
@@ -76,7 +76,7 @@ RUN yum upgrade -y \
           pkgconfig \
           systemd-devel \
           zlib-devel \
-          nc
```
kzys, line 79: Furthermore, do we really need to install `pkgconfig` and `-devel` packages? We don't build anything inside this container.

PettitWesley: So we started with this dockerfile: https://github.com/fluent/fluent-bit/blob/master/Dockerfile#L64

And then converted it to make it work on to amazon linux. I think there are some runtime dependencies needed. 

I am not an expert in this though and did it fairly hastily... making this image smaller has been on our backlog for a long time but we've never gotten to it. So thanks for helping.

I am fairly certain that `systemd-devel` is needed. I remember that there was an issue in the first version because I didn't include any runtime dependencies which broke the systemd/journald logs input plugin. Adding systemd to the image fixed it.

PettitWesley (approved):

Dockerfile
PettitWesley: I think `nc` can probably be removed. It was supposed to be part of a recommended healthcheck for FireLens but then we never published it. 

There's a very small chance that some agent functional tests were written assuming the command is there, and removing it might break those.

PettitWesley: If you think any of the other runtime dependencies can removed based on looking at the debian/distroless dockerfile let me know.

PettitWesley (approved):

PR #158, opened 2021-03-05
Add docs for aws-for-fluent-bit stable release

zhonghui12:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
 
 - [Versioning FAQ](#versioning-faq)
 - [Public Images](#public-images)
+    - [Using the stable tag](#using-the-stable-tag)
     - [Using SSM to find available versions](#using-ssm-to-find-available-versions)
     - [Using SSM Parameters in CloudFormation Templates](#using-ssm-parameters-in-cloudFormation-templates)
     - [Using image tags](#using-image-tags)
@@ -48,7 +49,13 @@ No. We continue to consume Fluent Bit from its main repository. We are not forki
 
 ### Public Images
 
-Each release updates the `latest` tag and adds a tag for the version of the image.
+Each release updates the `latest` tag and adds a tag for the version of the image. The `stable` tag is also available which marks a release as the latest stable version.
+
+#### Using the stable tag
+
+A `stable` tag can be trusted that it is the latest version in which there are no high impact bugs. A release may be marked as stable if the following rules are all met:
+* It has been out for at least 2 weeks
+* No bugs have been reported which we expect will have high impact for AWS customers. This means bugs in the components that are most frequently used by AWS customers, such as the AWS outputs of the tail input
 
 #### Using SSM to find available versions
 


README.md
```diff
@@ -48,7 +48,10 @@ No. We continue to consume Fluent Bit from its main repository. We are not forki
 
 ### Public Images
 
-Each release updates the `latest` tag and adds a tag for the version of the image.
+Each release updates the `latest` tag and adds a tag for the version of the image. A `stable` tag is also available which can be trusted that it is the latest version in which there are no high impact bugs. A release may be marked as stable if the following rules are all met:
+* It has been out for a period of time
+* The Fluent Bit version contained in the release is a patch version as the patch version indicates the release includes bug fixes
+* It has been no bugs reported which will have high impact for any of our customers
```
zhonghui12, line 55: Another idea is to create a subsection with title like "Using stable tag to find the latest version". But I think it may be a little inconsistent with "Using SSM to find available versions". Not sure which way is better. Any suggestions?

PettitWesley: I think actually it'd be best to create section right here, before "Using SSM to find available versions" which is titled

"Using the stable tag" or "Stable release tag". something like that.

That way it jumps out at you and is in the outline at the top. Otherwise it's very easy to miss this information.

README.md
```diff
@@ -48,7 +48,10 @@ No. We continue to consume Fluent Bit from its main repository. We are not forki
 
 ### Public Images
 
-Each release updates the `latest` tag and adds a tag for the version of the image.
+Each release updates the `latest` tag and adds a tag for the version of the image. A `stable` tag is also available which can be trusted that it is the latest version in which there are no high impact bugs. A release may be marked as stable if the following rules are all met:
+* It has been out for a period of time
```
PettitWesley, line 52: I feel this is too vague. May be we should just say that our criteria is at least 2 weeks?

README.md
```diff
@@ -48,7 +48,10 @@ No. We continue to consume Fluent Bit from its main repository. We are not forki
...
+* The Fluent Bit version contained in the release is a patch version as the patch version indicates the release includes bug fixes
```
PettitWesley, line 53: Let's remove this criteria. We will use it internally but let's not say it publicly. Because if we say it publicly we are admitting we don't trust the fluent bit upstream to release stable 1.x.0 versions.

README.md
```diff
@@ -48,7 +48,10 @@ No. We continue to consume Fluent Bit from its main repository. We are not forki
...
+* It has been no bugs reported which will have high impact for any of our customers
```
PettitWesley, line 54: Let's rephrase this a bit:

"No bugs have been reported which we expect will have high impact for AWS customers. This means bugs in the components that are most frequently used by AWS customers, such as the AWS outputs of the tail input."

PettitWesley (approved):

PR #169, opened 2021-03-26
Add sample configuration files

zhonghui12:
*Issue #, if available:*

*Description of changes:*
Add some common config files for Fluent Bit and provide additional options

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,6 +97,7 @@ COPY --from=aws-fluent-bit-plugins:latest /kinesis-streams/THIRD-PARTY \
     /kinesis-streams/LICENSE \
     /fluent-bit/licenses/kinesis/
 COPY AWS_FOR_FLUENT_BIT_VERSION /AWS_FOR_FLUENT_BIT_VERSION
+ADD ecs /ecs/
 
 COPY entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
--- a/ecs/amazon-es.conf
+++ b/ecs/amazon-es.conf
@@ -1,0 +1,14 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input-forward.conf
+@INCLUDE /ecs/ecs-metadata.conf
+
+[OUTPUT]
+    Name  es
+    Match *
+    Host  ${ES_ENDPOINT}
+    Port  443
+    Index ${INDEX}
+    Type  ${TYPE}
+    AWS_Auth On
+    AWS_Region ${ES_REGION}
+    tls   On
--- a/ecs/cloudwatch.conf
+++ b/ecs/cloudwatch.conf
@@ -1,0 +1,11 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input-forward.conf
+@INCLUDE /ecs/ecs-metadata.conf
+
+[OUTPUT]
+    Name   cloudwatch_logs
+    Match  *
+    region ${AWS_REGION}
+    log_group_name /aws/ecs/containerinsights/${ECS_CLUSTER}/application
+    log_stream_prefix ecs-
+    auto_create_group On
--- a/ecs/filter-parser.conf
+++ b/ecs/filter-parser.conf
@@ -1,0 +1,5 @@
+[FILTER]
+    Name  parser
+    Match *
+    Key_Name log
+    Parser ${PARSER_NAME}
--- a/ecs/firehose.conf
+++ b/ecs/firehose.conf
@@ -1,0 +1,9 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input-forward.conf
+@INCLUDE /ecs/ecs-metadata.conf
+
+[OUTPUT]
+    Name   kinesis_firehose
+    Match  *
+    region ${FIREHOSE_REGION}
+    delivery_stream ${DELIVERY_STREAM}
--- a/ecs/kinesis.conf
+++ b/ecs/kinesis.conf
@@ -1,0 +1,9 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input-forward.conf
+@INCLUDE /ecs/ecs-metadata.conf
+
+[OUTPUT]
+    Name   kinesis_streams
+    Match  *
+    region ${KINESIS_REGION}
+    stream ${STREAM}
--- a/ecs/s3.conf
+++ b/ecs/s3.conf
@@ -1,0 +1,14 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input-forward.conf
+@INCLUDE /ecs/ecs-metadata.conf
+
+[OUTPUT]
+    Name s3
+    Match *
+    bucket ${BUCKET}
+    region ${S3_REGION}
+    total_file_size 1M
+    upload_timeout 1m
+    use_put_object On
+    s3_key_format  /${ECS_CLUSTER}/${TASK_ID}/$TAG[0]/%Y/%m/%d/%H/%M/%S
+    s3_key_format_tag_delimiters .-
--- a/ecs/ecs-metadata.conf
+++ b/ecs/ecs-metadata.conf
@@ -1,0 +1,6 @@
+[FILTER]
+    Name record_modifier
+    Match *
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
--- a/ecs/input-forward.conf
+++ b/ecs/input-forward.conf
@@ -1,0 +1,3 @@
+[INPUT]
+    Name forward
+    unix_path /var/run/fluent.sock
--- a/ecs/service.conf
+++ b/ecs/service.conf
@@ -1,0 +1,2 @@
+[SERVICE]
+    Grace 30


ecs/firehose.conf
```diff
@@ -0,0 +1,17 @@
+[INPUT]
+    Name forward
+    unix_path /var/run/fluent.sock
+
+[FILTER]
+    Name record_modifier
+    Match *
+    Record ec2_instance_id true
```
PettitWesley, line 8: This will add a key the logs like this
```
"ec2_instance_id": "true",
```

You need to use the AWS Filter to add the instance ID.

ecs/firehose.conf
```diff
@@ -0,0 +1,17 @@
...
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
+
+[OUTPUT]
+    Name   firehose
```
PettitWesley, line 14: use the new high performance C output

ecs/firehose.conf
```diff
@@ -0,0 +1,17 @@
+[INPUT]
```
PettitWesley, line 1: One thing I forgot to mention, we should add a service section with grace period set to 30:

```
[SERVICE]
     Grace 30
```

zhonghui12:
built the image with this code and it succeeded

ecs/firehose.conf
```diff
@@ -0,0 +1,24 @@
...
+    Grace 30
+
+[INPUT]
+    Name forward
+    unix_path /var/run/fluent.sock
+
+[FILTER]
+    Name aws
+    Match *
+    ec2_instance_id true
```
PettitWesley, line 11: Sorry... I just realized we shouldn't add this. Because on Fargate it won't work, there's not instance ID there. (Also, this plugin uses the EC2 metadata service, which many ECS customers disable, so its not safe to assume it will always be there). So we can't have it in a built-in generic config for ECS.

PettitWesley (approved):
Aside from the new comment about realizing that we can't add the EC2 instance fields, this looks good

ecs/amazon-es.conf
```diff
@@ -0,0 +1,21 @@
...
+    Match *
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
+
+[OUTPUT]
+    Name  es
+    Match *
+    AWS_Region ${ES_REGION}
+    AWS_STS_Endpoint ${ENDPOINT}
```
PettitWesley, line 19: the endpoint is for elasticsearch. Its the `Host` field. Also this config will not actually work to send logs to Amazon ES. Please see and copy this: https://docs.fluentbit.io/manual/pipeline/outputs/elasticsearch##aws-es

PettitWesley: also, from my doc, you are missing the INDEX and TYPE env vars

PettitWesley: also, I just updated the doc, the endpoint should be `ES_ENDPOINT` for the env var

ecs/cloudwatch.conf
```diff
@@ -0,0 +1,21 @@
...
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
+
+[OUTPUT]
+    Name   cloudwatch_logs
+    Match  *
+    region us-east-1
+    log_group_name fluent-bit-cloudwatch
+    log_stream_prefix from-fluent-bit-
```
PettitWesley, line 20: these are not the log group and stream names that we want. 

For now let's do:
```
log_group_name /aws/ecs/containerinsights/${ECS_CLUSTER}/application
log_stream_prefix ecs-
```

ecs/amazon-es.conf
zhonghui12:
```
    Index ${INDEX}
    Type  ${TYPE}
```
are these enough for INDEX and TYPE env vars?

ecs/filter-parser.conf
```diff
@@ -0,0 +1,21 @@
+[SERVICE]
+    Parsers_File /fluent-bit/parsers/parsers.conf
+    Grace 30
+
+[INPUT]
+    Name forward
+    unix_path /var/run/fluent.sock
+
+[FILTER]
+    Name ${PARSER_NAME}
```
PettitWesley, line 10: this is the name of the filter plugin, which is `parser`

ecs/filter-parser.conf
```diff
@@ -0,0 +1,21 @@
...
+    Match *
+    Key_Name log
+    Parser json
```
PettitWesley, line 13: this is where you add PARSER_NAME

ecs/filter-parser.conf
```diff
@@ -0,0 +1,21 @@
+[SERVICE]
+    Parsers_File /fluent-bit/parsers/parsers.conf
+    Grace 30
+
+[INPUT]
+    Name forward
+    unix_path /var/run/fluent.sock
```
PettitWesley, line 7: for this config, its just the parser filter, nothing else

ecs/filter-parser.conf
```diff
@@ -0,0 +1,21 @@
...
+    Key_Name log
+    Parser json
+    Reserve_Data True
+
+[FILTER]
+    Name record_modifier
+    Match *
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
```
PettitWesley, line 21: for this config, its just the parser filter, nothing else

PettitWesley: the idea is that you use this config along with one of the others. The other one will contain the input and output. This one just adds filter.

ecs/firehose.conf
```diff
@@ -0,0 +1,19 @@
...
+    Match *
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
+
+[OUTPUT]
+    Name   kinesis_firehose
+    Match  *
+    region ${FIREHOSE_REGION}
+    delivery_stream ${STREAM}
```
PettitWesley, line 19: should be DELIVERY_STREAM

PettitWesley: my doc was wrong, I just updated it. The name is supposed to match the plugin config key name

ecs/kinesis.conf
```diff
@@ -0,0 +1,19 @@
...
+    Match *
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
+
+[OUTPUT]
+    Name   kinesis_streams
+    Match  *
+    region ${KINESIS_REGION}
+    stream ${DELIVERY_STREAM}
```
PettitWesley, line 19: just STREAM, see note above

ecs/s3.conf
```diff
@@ -0,0 +1,19 @@
...
+    Match *
+    Record ecs_cluster ${ECS_CLUSTER}
+    Record ecs_task_arn ${ECS_TASK_ARN}
+    Record ecs_task_definition ${ECS_TASK_DEFINITION}
+
+[OUTPUT]
+    Name s3
+    Match *
+    bucket ${BUCKET}
+    region ${S3_REGION}
```
PettitWesley, line 19: please use the settings recommended for scenario with no persistent disk as shown here: https://docs.fluentbit.io/manual/pipeline/outputs/s3#using-s3-without-persisted-disk

also, we need to set S3 key format. I look forward to seeing you idea for that... it should be something that includes the ECS Metadata..

ecs/amazon-es.conf
PettitWesley: Yes

ecs/s3.conf
zhonghui12: so maybe we need support ecs metadata in s3_key_format

ecs/firehose.conf
hossain-rayhan: Can two different configs be an option? For EC2 and Fargate?

PettitWesley: Since customers will be able to specify multiple configs, we could have a separate config with just the EC2 filter. 

Then if you are running on EC2, you take your base config with the input and output, and you also add the ec2 filter config.

hossain-rayhan: Thats also nice.

ecs/s3.conf
```diff
@@ -0,0 +1,23 @@
...
+    total_file_size 1M
+    upload_timeout 1m
+    use_put_object On
+    s3_key_format  /${ECS_CLUSTER}/${ECS_TASK_ID}/%Y/%m/%d/%H/%M/%S
```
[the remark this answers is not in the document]
PettitWesley: Yeah that's the public doc for FireLens V1, with our new feature it will change to what I noted in my comment, and these configs are for the new feature...

Dockerfile
```diff
@@ -54,6 +54,18 @@ RUN install bin/fluent-bit /fluent-bit/bin/
...
+COPY ecs/cloudwatch.conf \
+     /ecs/
+COPY ecs/kinesis.conf \
+     /ecs/
+COPY ecs/amazon-es.conf \
+     /ecs/
+COPY ecs/s3.conf \
+     /ecs/
+COPY ecs/filter-parser.conf \
+     /ecs/
```
hossain-rayhan, line 68: Wondering can't we just copy the whole `/ecs` folder in one command?

zhonghui12: good point. Let me explore on it

hossain-rayhan:
Hey @PettitWesley @zhonghui12 I am not sure but wondering, I think the SERVICE, INPUT and FILTER configurations are fully or at least mostly similar. Can't we just put them in a common file and import from all other plugin specifc configs?

zhonghui12:
> Hey @PettitWesley @zhonghui12 I am not sure but wondering, I think the SERVICE, INPUT and FILTER configurations are fully or at least mostly similar. Can't we just put them in a common file and import inside all other plugin specifc configs?

I am not sure about it. But I think it may be more clear and complete for users to see different config files. I am okay with these two ways

PettitWesley:
> Can't we just put them in a common file and import inside all other plugin specific configs?

@zhonghui12 @hossain-rayhan Good idea we could use an `@INCLUDE` to import a common config. 

Also... actually I am wondering if we want to explicitly separate the input and output. That way users can use multiple outputs like this by providing a list of config files like this:
* /ecs/input/stdout.conf
* /ecs/output/s3.conf
* /ecs/output/kinesis.conf

With the current design, a user can only select one of these built in config files (if they choose two, the input definition will be duplicated). If we change it to separate out the input, then they can easily add as many as needed.

hossain-rayhan:
> > Can't we just put them in a common file and import inside all other plugin specific configs?
> 
> @zhonghui12 @hossain-rayhan Good idea we could use an `@INCLUDE` to import a common config.
> 
> Also... actually I am wondering if we want to explicitly separate the input and output. That way users can use multiple outputs like this by providing a list of config files like this:
> 
> * /ecs/input/stdout.conf
> * /ecs/output/s3.conf
> * /ecs/output/kinesis.conf
> 
> With the current design, a user can only select one of these built in config files (if they choose two, the input definition will be duplicated). If we change it to separate out the input, then they can easily add as many as needed.

Yea, this would be more powerful. Also, we need to think about making it simple for customers. We will surely provide some full config files with INPUT, FILTER, SERVICE and OUTPUT, right?

PettitWesley:
@hossain-rayhan Yeah I think its a good idea... I had the same thought myself. We should provide "full" configs for common destinations, and also separate configs for each component. 

Or may be we just do full configs and then if you want to add multiple outputs you must create your own. Let's discuss this sometime.

For now @zhonghui12 let's just stick with Rayhan's first suggestion to use `@INCLUDE` and keep these configs "full" as we originally planned.

Dockerfile
hossain-rayhan: Also, I prefer to move all these files under `config/ecs/`.

PettitWesley: I don't like it... it makes the path longer. Remember that folks will specify the full path in our API- I want it to be as short as possible

hossain-rayhan: Thinking if we consider adding default configs for EKS and other platforms also. In that case all the configs under `/config` folder might be considered as good design.

PettitWesley: For EKS there is not a strong need I think, config maps exist and are super easy to use. 

This work is just for our new project...

hossain-rayhan: If so, why do we need to mention `/ecs`. We may put everything under `/config` folder.

hossain-rayhan: I am OK with either of them. Let's move with our original plan (only `/ecs`).

ecs/cloudwatch.conf
```diff
@@ -0,0 +1,11 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input.conf
+@INCLUDE /ecs/filter-record.conf
```
PettitWesley, line 3: let's call this, `ecs-metadata.conf`

hossain-rayhan (approved):
Thanks. LGTM.

ecs/s3.conf
```diff
@@ -0,0 +1,14 @@
...
+
+[OUTPUT]
+    Name s3
+    Match *
+    bucket ${BUCKET}
+    region ${S3_REGION}
+    total_file_size 1M
+    upload_timeout 1m
+    use_put_object On
+    s3_key_format  /$TAG[0]/${TASK_ID}/%Y/%m/%d/%H/%M/%S
```
PettitWesley, line 13: S3 key should also have cluster

I think it should go from least specific, to most specific, so:

`/cluster/task ID/container name`

ecs/cloudwatch.conf
```diff
@@ -0,0 +1,11 @@
+@INCLUDE /ecs/service.conf
+@INCLUDE /ecs/input-forward.conf
+@INCLUDE /ecs/ecs-metadata.conf
+
+[OUTPUT]
+    Name   cloudwatch_logs
+    Match  *
+    region us-east-1
```
PettitWesley, line 8: I just realized region needs to be parametrized too...

Let's say that the init process also adds an env var `AWS_REGION`. Which you can use here. 

For CW, we assume folks want to send logs to the same region that they are running tasks in. For other outputs like firehose, while they should use the same region, they don;t have to, hence we need to make the region come from ther user.

PettitWesley (approved):

PR #189, opened 2021-06-02
fix wrong ssm parameter for stable image issue

zhonghui12:
*Issue #, if available:*

*Description of changes:*
Fix wrong ssm parameter for stable image issue by adding a check for the value of existing ssm parameter.


hossain-rayhan (commented):
Do we need to do anything else for the verification of stable SSM?

PettitWesley (approved):

zhonghui12:
> Do we need to do anything else for the verification of stable SSM?

For our current logic, we do verify-ssm by checking its region and pulling the image. Do we need to do anything else? This PR is to resolve the issue of not updating parameter value of stable image ssm parameter. This is because stable image is not like other images. It will always have the same name in ssm parameter so we need to use the value to see if it has been updated. (so maybe we already add one check here for stable image)

hossain-rayhan (approved):

PR #199, opened 2021-06-28
log into dockerhub in sync tasks only when we are in us-west-2

zhonghui12:
*Issue #, if available:*

*Description of changes:*
Our sync tasks fail in most of regions and this is because we added dockerhub login in our verification step. In this case, we should only call this function when we are in us-west-2.

### DIFF
--- a/buildspec_sync.yml
+++ b/buildspec_sync.yml
@@ -60,8 +60,8 @@ phases:
       - export AWS_SECRET_ACCESS_KEY=
       - export AWS_SESSION_TOKEN=
 
-      # Verify the publishing on Public ECR and Dockerhub
-      - './scripts/publish.sh cicd-verify stable public-repo'
+      # Verify the publishing on Public ECR and Dockerhub when AWS_REGION is us-west-2
+      - './scripts/publish.sh cicd-verify stable ${AWS_REGION}'
 artifacts:
   files:
     - '**/*'
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -724,8 +724,10 @@ if [ "${1}" = "cicd-verify" ]; then
 			verify_ecr ${region} ${classic_regions_account_id} true
 		done
 	elif [ "${2}" = "stable" ]; then
-		verify_dockerhub stable
-		verify_public_ecr stable
+		if [ "${3}" = "us-west-2" ]; then
+			verify_dockerhub stable
+			verify_public_ecr stable
+		fi
 	else
 		verify_ecr "${2}" ${classic_regions_account_id}
 	fi


buildspec_sync.yml
```diff
@@ -61,7 +61,7 @@ phases:
       - export AWS_SESSION_TOKEN=
 
       # Verify the publishing on Public ECR and Dockerhub
-      - './scripts/publish.sh cicd-verify stable public-repo'
+      - './scripts/publish.sh cicd-verify stable public-repo ${AWS_REGION}'
```
hossain-rayhan, line 64: Wondering, where are we using the 3rd variable `public-repo`?

zhonghui12: it is just a sign that we are verify our public repo. Do we need to remove it? (Don't want to make it confused for other engineers..

hossain-rayhan: Unnecessary variable. Better to remove it. We do have the comment. 

Also, modify the comment that this will only occur when the region is `us-west-2`

zhonghui12: Update

buildspec_sync.yml
```diff
@@ -60,8 +60,8 @@ phases:
       - export AWS_SECRET_ACCESS_KEY=
       - export AWS_SESSION_TOKEN=
 
-      # Verify the publishing on Public ECR and Dockerhub
-      - './scripts/publish.sh cicd-verify stable public-repo'
```
PettitWesley, line 64: the existing `public-repo` argument isn't used anywhere?

hossain-rayhan (approved):
Looks good! Thanks.

scripts/publish.sh
```diff
@@ -724,8 +724,10 @@ if [ "${1}" = "cicd-verify" ]; then
 			verify_ecr ${region} ${classic_regions_account_id} true
 		done
 	elif [ "${2}" = "stable" ]; then
-		verify_dockerhub stable
-		verify_public_ecr stable
+		if [ "${3}" = "us-west-2" ]; then
+			verify_dockerhub stable
+			verify_public_ecr stable
+		fi
```
PettitWesley, line 730: I see the same two functions being called with the same arguments- how does this change anything?

buildspec_sync.yml
zhonghui12: Previously, we need it as we need 3 arguments here to check if we are in which case. But now, I've deleted it and add the AWS_REGION as another arg

scripts/publish.sh
zhonghui12: several days ago, we has this commit: https://github.com/aws/aws-for-fluent-bit/commit/5eb9a3fb37ca6bc550b76587295fefa6f0397b29. It adds dockerhub login in `verify_dockerhub()`. In this case, we can only call `verify_dockerhub()` in our main account. In our other accounts like BAH, it cannot log into our dockerhub as it doesn't have the creds. Therefore, we need to add a check here that we only call it when we are in `us-west-2`

PettitWesley: Oh I see. You are only doing these steps in us-west-2 because that the only stage where we verify public ECR/Dockerhub?

PettitWesley: So basically, previously we had a bug because we were verifying docker hub in all regions, even though we only need to do it once??

PettitWesley (approved):

PR #209, opened 2021-07-26
firelens example for using multiline filter

zhonghui12:
*Description of changes:*
Add an example for using multiline filter with Fluent Bit. It is from https://docs.fluentbit.io/manual/pipeline/filters/multiline-stacktrace

### DIFF
--- a/multiline/parsers_multiline.conf
+++ b/multiline/parsers_multiline.conf
@@ -1,0 +1,16 @@
+[MULTILINE_PARSER]
+    name          multiline-regex-test
+    type          regex
+    # Regex rules for multiline parsing
+    # ---------------------------------
+    #
+    # configuration hints:
+    #
+    #  - first state always has the name: start_state
+    #  - every field in the rule must be inside double quotes
+    #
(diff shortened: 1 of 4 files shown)


multiline/parsers_multiline.conf
```diff
@@ -0,0 +1,16 @@
...
+    #
+    # configuration hints:
+    #
+    #  - first state always has the name: start_state
+    #  - every field in the rule must be inside double quotes
+    #
+    # rules   |   state name   | regex pattern                   | next state name
+    # --------|----------------|--------------------------------------------------
+    rule         "start_state"   "/(Dec \d+ \d+\:\d+\:\d+)(.*)/"  "cont"
+    rule         "cont"          "/^\s+at.*/"                     "cont"
```
hossain-rayhan, line 15: I am not sure but the state_name and next_state_name are same. Is that OK?

zhonghui12: yeah It works. It is from upstream doc

zhonghui12:
Close this PR as the example needs to be uploaded in another repo.

PR #221, opened 2021-08-12
Update image version check logic in build stage

zhonghui12:
Signed-off-by: Zhonghui Hu <zh0512xx@gmail.com>

*Issue #, if available:*

*Description of changes:*
Update version check logic to let it check with all previous tags so that only new version can be released.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -350,8 +350,11 @@ verify_ecr() {
 }
 
 check_image_version() {
-	if [ "${1}" = "${2}" ]; then
-		echo "Accidental release: latest image versions from dockerhub and from github source file match."
+	export DOCKER_CLI_EXPERIMENTAL=enabled
+	# check if we can get the image information in dockerhub; if yes, the exit status should be 0
+	docker manifest inspect amazon/aws-for-fluent-bit:${1} > /dev/null
+	if [ "$?" = "0" ]; then
+		echo "Accidental release: current image version from github source file match a previous version from dockerhub."
 		exit 1
 	fi
 }
@@ -792,5 +795,5 @@ if [ "${1}" = "cicd-verify-ecr-image-scan" ]; then
 fi
 
 if [ "${1}" = "cicd-check-image-version" ]; then
-	check_image_version ${AWS_FOR_FLUENT_BIT_VERSION} ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
+	check_image_version ${AWS_FOR_FLUENT_BIT_VERSION} 
 fi


scripts/publish.sh
```diff
@@ -350,8 +350,9 @@ verify_ecr() {
 }
 
 check_image_version() {
-	if [ "${1}" = "${2}" ]; then
-		echo "Accidental release: latest image versions from dockerhub and from github source file match."
+	docker_hub_image_exists=$(curl -s -S https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/${1} | jq -r '.text')
+	if [ "${docker_hub_image_exists}" = "null" ]; then
+		echo "Accidental release: current image version from github source file match previous version from dockerhub."
```
hossain-rayhan, line 355:
```suggestion
		echo "Accidental release: current image version from github source file matches a previous version from dockerhub."
```

scripts/publish.sh
```diff
@@ -350,8 +350,11 @@ verify_ecr() {
 }
 
 check_image_version() {
-	if [ "${1}" = "${2}" ]; then
-		echo "Accidental release: latest image versions from dockerhub and from github source file match."
+	docker_hub_image_exists=$(curl -s -S https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/${1} | jq -r '.text')
```
PettitWesley, line 353: hmmm. How robust is this? How likely is it to break if DockerHub changes the output of their API? Do they have official documentation on this? We need to be careful about writing code which works in 2021 but 2 years from now might break in unexpected ways because DockerHub changes.

I think it is acceptable to rely on them as long as our check is fairly robust. For this, what happens if they remove or change how this text field works? Will it negatively impact us? Or will it just block the pipeline. If it only causes the script to fail, then I think its fine. The release will fail 2 years from now if they make a change but we will fix it quickly and customers will not be impacted.

What do you think?

zhonghui12: yeah you are right. I will check to see if there are any official docs. I planned to do something like 
```
if docker_hub_image_exists != *"not found"*
```
But yes, I think we need to do it if it is really robust.

zhonghui12: After a deeper investigation, I think the response might depend on the dockerhub API. We cannot ensure if it will change after it is updated.  I think we might be able to use `docker manifest inspect` to check it. If there is an existing image, then it will return `success`. But if doesn't exist, the command will fail and we then continue the build step.

PettitWesley: > docker manifest inspect

Is this a remote call or it relies on the image being locally present?

zhonghui12: A remote call. We need to login into the dockerhub first

zhonghui12: A remote call. I plan to inspect to our images on dockerhub. But I find we don't need to log into our dockerhub first. Maybe it is because our repo is public

hossain-rayhan: Thats a good point- this API is not official and it will affect our logic if they change the output. To me the list tag seems good for now as we have a small number of tags. In future, if Docker make this API official, we can take a dependency on this.

PettitWesley (approved):

hossain-rayhan (approved):

PR #238, opened 2021-09-22
Add `tini` as entrypoint

### LINKED ISSUE #174: Exit Code 137 for the v2.12.0
We are triggering fargate tasks with aws cli. For the v2.12.0 fluentbit log router container fails because the container occasinally exits with exit code 137 when the main app container in the task finishes.

I have seen similar issue -> https://github.com/aws/aws-for-fluent-bit/issues/11

If you need more questions, please let me know

f0o:
*Issue #, if available:* Potentially Fixes #174

*Description of changes:*

This patch removed the obscure exit-code 137 for us.

I suspect that fluentbit does not handle the signals correctly as PID1. Thus the runtime may opt to SIGKILL it resulting in 137 exitcode.

`tini` is a tiny process manager that handles and forwards any and all signals in accordance to POSIX.

See more about `tini` at https://github.com/krallin/tini

As of submission we have not gotten any alarm about non-zero exitcodes for over 4 hours. Without the patch we would see this as frequently as every hour.

I will keep this as a draft until more time passes and I can fully state that it is fixed (next week?)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,8 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:latest as builder
 ENV FLB_VERSION 1.8.6
 # branch to pull parsers from in github.com/fluent/fluent-bit-docker-image
 ENV FLB_DOCKER_BRANCH 1.8
+# tini version to install
+ENV TINI_VERSION 0.19.0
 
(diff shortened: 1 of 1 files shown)


PettitWesley:
@f0o How have your tests gone? Do you still want us to consider this patch?

PR #247, opened 2021-09-29
Fix the overwritten issue for latest tag in classic region

zhonghui12:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -230,8 +230,10 @@ sync_latest_image() {
 		fi
 	done
 
-	create_manifest_list ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit "latest" ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -230,8 +230,10 @@ sync_latest_image() {
 		fi
 	done
 
-	create_manifest_list ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit "latest" ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
-	create_manifest_list ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB} ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
+	if [ "${account_id}" != '906394416424' ]; then
```
PettitWesley, line 233: there is a variable for this account ID I think... if not let's add one so its more clear to future maintainers why we added this code.

PettitWesley (approved):

PR #260, opened 2021-10-13
Bump version to 2.21.0

PettitWesley:

This build relies on a branch on my fork which contains 1.8.8 + the same commits we cherry-picked into the last release for the TLS and mbedtls fixes: https://github.com/PettitWesley/fluent-bit/tree/1_8_8_patch


Dockerfile
```diff
@@ -41,9 +41,9 @@ ENV PATH ${PATH}:/home/.gimme/versions/go1.17.linux.arm64/bin:/home/.gimme/versi
 RUN go version
 
 WORKDIR /tmp/fluent-bit-$FLB_VERSION/
-RUN git clone https://github.com/zhonghui12/fluent-bit.git /tmp/fluent-bit-$FLB_VERSION/
+RUN git clone https://github.com/PettitWesley/fluent-bit.git /tmp/fluent-bit-$FLB_VERSION/
```
hossain-rayhan, line 44: Wondering why do we need to change this everytime. Can't we create a patch from our aws repo?

matthewfala: Zhonghui explained the issue last week to me when I shadowed her. this is just a fork of the fluent bit v1.8.8 + some extra commits that we have added. We could potentially do the patching/ cherry picking of commits directly in docker with commits to be cherry-picked along with the source repo listed in some file, if that's what you're thinking though.

hossain-rayhan: I was thinking rather than creatinng the patch on different repos `PettitWesley/patch` or `zhonghui12/patch`, can we do `aws/patch`. 
Not a big deal.

hossain-rayhan (approved):

Dockerfile
matthewfala: O i c. People may think that we are forking fluent bit.

I actually would prefer to have a list of commits to be cherry picked in a file on the repo and for those commits to be added on top of the fluent/fluent-bit.git automatically by Dockerfile. That way the cherrypicked commits are explicitly listed and highly visible in our git repo and pr diff. What do you think? (That is, if we continue to have custom builds with each version of fluent bit)

matthewfala: Actually, merge conflicts will be a nightmare. Maybe the dockerfile could throw an error if any merge conflict is detected. The current solution might be better.

hossain-rayhan: Can't disagree. Seems to be a good idea.

PettitWesley: > can we do aws/patch.

This would require an official AWS repo which would be an official AWS fork of fluent bit. We could do that, but it has public relations concerns. Folks will think we may be beginning a long term plan to diverge from upstream. 

I don't expect that building from our own patches should be a common thing... hopefully soon we can get everything we need in upstream...

Issue #270, opened 2021-11-21
Create an env variable for custom fluent-bit.conf path


thpham:
Hello,

could it be possible to create an environment variable in the [`entrypoint.sh`](https://github.com/aws/aws-for-fluent-bit/blob/mainline/entrypoint.sh#L3) that is used as the main CMD of the Dockerfile, so that we could define alternative path for the `fluent-bit.conf` ?

Thank you

PettitWesley:
@thpham Can you explain your use case in more detail? There may be simpler ways to achieve what you desire without changing our image. 

For example, you can use the @INCLUDE command in the main config file to import another config file: https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/configuration-file

If you are using FireLens, its easy to add a new config file: https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/config-file-type-file

thpham:
@PettitWesley , sure. In fact, this request is not directly link with your image, but this change would help me to use your image as a sidecar container in the [bank-vault operator](https://banzaicloud.com/docs/bank-vaults/overview/) to tail the logs to Cloudwatch or S3.  One problem is that they first supported `fluentd` config filename convention and they hardcoded this filename in their operator (which is far from ideal). But the operator allows to inject variables to the sidecar, so that it could be used to point to the correct configuration filename.

so that means my problem could be solve, either from your side or from them by avoiding to hardcode the filename. btw I did a [commit](https://github.com/thpham/bank-vaults/commit/506ffdcc9bba5fcafe5333fe31c06dfa3d8f5773) to express the idea.

Nevertheless, I think my request could be also beneficial for your image in other contexts and , correct me if I'm wrong, this change seems minimal to me.

What do you think ?

PettitWesley:
I see. So your specific use case is very specific. 

> I think my request could be also beneficial for your image in other contexts and , correct me if I'm wrong, this change seems minimal to me.

Yea I do think it's possibly generally useful, and it doesn't hurt to have it. I'm okay with it. Let's see what the rest of my team thinks.

matthewfala:
One potential problem I see is that the custom config file is not copied into docker, so I'm not sure how this environment variable would actually be useful, unless there is a bunch of config files already included in our image which the user can select from (I don't think this is the case).

[...]

PR #271, opened 2021-11-21
add FLUENT_BIT_CONFIG_FILE env variable to `entrypoint.sh`

thpham:
This PR relates to following issue: https://github.com/aws/aws-for-fluent-bit/issues/270

This would allow us to define custom path for the `fluent-bit.conf` in situation where we cannot touch k8s deployment manifest and mounted configmap. (Ex.: when they are wrapped by an operator that inject sidecar containers).

Thanks for your review.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,3 +1,6 @@
 echo -n "AWS for Fluent Bit Container Image Version "
 cat /AWS_FOR_FLUENT_BIT_VERSION
-exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so -c /fluent-bit/etc/fluent-bit.conf
+
+FLUENT_BIT_CONFIG_FILE=${FLUENT_BIT_CONFIG_FILE:-'/fluent-bit/etc/fluent-bit.conf'}
+
+exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so -c ${FLUENT_BIT_CONFIG_FILE}


PettitWesley:
Please dive deep into why you need this in the linked issue.

I'm not against this change if you can explain why its truly necessary. I know that the Fluentd container images have a similar env var set up for their config.

entrypoint.sh
```diff
@@ -1,3 +1,6 @@
 echo -n "AWS for Fluent Bit Container Image Version "
 cat /AWS_FOR_FLUENT_BIT_VERSION
-exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so -c /fluent-bit/etc/fluent-bit.conf
+
+AWS_FOR_FLUENT_BIT_CONFIGFILE=${AWS_FOR_FLUENT_BIT_CONFIGFILE:-'/fluent-bit/etc/fluent-bit.conf'}
```
PettitWesley, line 4: I think there should be an underscore between CONFIG and FILE to make it all consistent, so `AWS_FOR_FLUENT_BIT_CONFIG_FILE`

PettitWesley:
@aws/aws-firelens I'm okay with this change but want to see what others think.

See the discussion in #270

entrypoint.sh
hossain-rayhan: I prefer it to be `FLUENT_BIT_CONFIG_FILE` as aws-for-fluent-bit doesn't have a different config.

thpham: @hossain-rayhan  I agree ! , I'll do the change !

thpham:
In the related issue ticket, it has been decided that I should instead create a custom image for this niche use case. and that PR might not comply with long term vision or security aspects of the maintainers.
I'll close this PR. Thanks for review anyway.

PR #307, opened 2022-03-03
Added debug dockerfile and instructions for use

PettitWesley:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -66,9 +66,13 @@ Fluent Bit is efficient and performant, but it isn't magic. Here are common/norm
 
 Please see the Fluent Bit buffering documentation: https://docs.fluentbit.io/manual/administration/buffering-and-storage
 
-#### Testing for real memory leaks
+#### Testing for real memory leaks using Valgrind
 
 If you do think the cause of your high memory usage is a bug in the code, you can optionally help us out by attempting to profile Fluent Bit for the source of the leak. To do this, use the [Valgrind](https://valgrind.org/) tool.
+
+There are some caveats when using Valgrind. Its a debugging tool, and so it significantly reduces the performance of Fluent Bit. It will consume more memory and CPU and generally be slower when debugging. Therefore, Valgrind images may not be safe to deploy to prod. 
+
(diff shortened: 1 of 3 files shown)


troubleshooting/debugging.md
```diff
@@ -89,11 +93,22 @@ COPY --from=builder /fluent-bit /fluent-bit
 CMD valgrind --leak-check=full /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
 ```
 
-There are some caveats when using Valgrind. Its a debugging tool, and so it significantly reduces the performance of Fluent Bit. It will consume more memory and CPU and generally be slower when debugging. Therefore, Valgrind images may not be safe to deploy to prod. 
+##### Option 2: Debug Build (More robust)
+
+The best option, which is most likely to catch any leak or segfault is to create a fresh build of the image using the [`Dockerfile.debug`](https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile.debug) in AWS for Fluent Bit. Check out the git tag for the version that saw the problem, and then build this dockerfile with the `make debug` target. Make sure the `FLB_VERSION` at the top of the `Dockerfile.debug` is set to the same version as the main Dockerfile for that tag. This will create a fresh build with debug mode and valgrind support enabled, which gives the highest chance that Valgrind will be albe to produce useful diagnostic information about the issue.
```
DrewZhang13, line 98: Great to see we have this option. Just some thinking: 
Do we want to make the benefit and actions separately. like

one paragraph for talking what's this options and why it's recommended. 

another paragraph or just a list of items how customer should use this options. Like;
1. check out the github version
2. make sure fluent bit version is matching
3. run `make debug`

finally what's the output or expectation after the actions above. 

This is open to discuss.

DrewZhang13 (commented):
Overall LGTM, glad we/customers have this options now for debugging!

troubleshooting/debugging.md
PettitWesley: Addressed, thanks

DrewZhang13 (approved):

PR #310, opened 2022-03-09
set more buffer for logs sent to corresponding destination

zhonghui12:
*Issue #, if available:*

*Description of changes:*
Based on what I've observed, we need to set more buffer for logs to be sent to the destination. Otherwise, we will see more log loss than it should be.


matthewfala (approved):
LGTM

PR #320, opened 2022-03-25
Opensource devtools guide

matthewfala:
*Issue #, if available:*

*Description of changes:*


troubleshooting/devtools.md
```diff
@@ -0,0 +1,188 @@
+# Dev Tools Guide
+
+## VS Code Remote Setup
+**VSCode Remote**  
+After installing [VS Code](https://code.visualstudio.com/Download), the following tutorial can be used to connect Visual Studio Code to a remote Linux system or server. <https://code.visualstudio.com/docs/remote/ssh>
```
zhonghui12, line 5: I wonder prefer to add link to a string rather than show link directly. like [Remote Development using SSH](https://code.visualstudio.com/docs/remote/ssh)

troubleshooting/devtools.md
```diff
@@ -0,0 +1,188 @@
...
+1. Sets style guides
+    - Indent style
+    - Ruler set to 90 characters from left
+2. Adds build scripts used by VS Code debugging
+3. Adds build tasks used by VS Code debugging
+4. Adds launch configuration `launch.json` which sets up vscode to build and run Fluent Bit
+5. External plugins folder, to auto build, attach, and run Go plugins if needed
+6. Fluent Bit config files that are connected to the Fluent Bit launch tasks
+
+If you wish to develop aws-for-fluent-bit go plugins, we recommend you follow the readme in the external plugins folder to configure these repositories.
```
zhonghui12, line 55: nit: change readme to full name README.md and add link to aws-for-fluent-bit go plugins: https://github.com/aws/aws-for-fluent-bit#plugins

troubleshooting/devtools.md
```diff
@@ -0,0 +1,188 @@
...
+To run the unit tests in VSCode, first configure CMake
+```
+cmake -DFLB_DEV=On -DFLB_TESTS_RUNTIME=On -DFLB_TESTS_INTERNAL=On ../
+make
+```
+
+At the blue ribbon bar of VS Code, you should see a button that says Build [all]. Press that button.
+When complete, click the text to the right of the debug and play button, which allows us to select our debug target. Click the unit test you would like to run. Then press the debug button.
+
+If you insert breakpoints in you test you will notice that these breakpoints, even though we are running in debug mode ***will not be hit!!!***
```
zhonghui12, line 96: As it is an official doc, I think we might not want it to be too over. Just one mark is fine. like **will not be hit!***

troubleshooting/resources/vscode/fluent-bit-config/fluent-bit-cloudwatch_logs.conf
```diff
@@ -0,0 +1,25 @@
...
+     Name                dummy
+
+[OUTPUT]
+     Name                cloudwatch_logs
+     Match               *
+     log_stream_prefix   my_prefix
+     log_group_name      my_group_name
+     auto_create_group   true
+     region              us-west-2
+     #log_key            log
```
zhonghui12, line 25: we could add a comment about this line as it is not that intuitive

troubleshooting/resources/vscode/scripts/list-fluent-bit-plugin-args.sh
```diff
@@ -0,0 +1,10 @@
+
```
zhonghui12, line 1: nit: extra line

zhonghui12 (approved):
Super helpful! Left a few small comments. Thanks

troubleshooting/devtools.md
matthewfala: Actually I am referring to another readme, the one included in the external-plugins folder. Changed to refer to that doc.

matthewfala:
Great comments! Thank you for taking a thorough look at it!
Made all the appropriate changes.

PR #322, opened 2022-03-30
Bump to 2.23.3

matthewfala:
Signed-off-by: Matthew Fala <falamatt@amazon.com>

*Issue #, if available:*

*Description of changes:*


CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
...
+* Fluent Bit [1.8.15](https://fluentbit.io/announcements/v1.8.15/)
+* Amazon CloudWatch Logs for Fluent Bit 1.7.0
+* Amazon Kinesis Streams for Fluent Bit 1.9.0
+* Amazon Kinesis Firehose for Fluent Bit 1.6.1
+
+Same as `2.23.2`, this release includes the following fix for AWS customers that we are working on getting accepted upstream:
+* Bug - Resolve IMDSv1 fallback error introduced in 2.21.0 [aws-for-fluent-bit:259](https://github.com/aws/aws-for-fluent-bit/issues/259)
+
+Note:
+* This release includes several core fixes added in Fluent Bit 1.8.15 including enhanced networking thread safety, and keep_alive connection error handling.
```
PettitWesley, line 14: I don't think we need this honestly. I know there isn't anything new except the bump to 1.8.15, but that's already called out at the top with a link to the release notes. We don't have any specific data indicating that these fixes affect AWS customers, we're just putting them out in case they do. THus, I think this call out is in some sense misleading since it implies we have reason to call out these fixes.

matthewfala: Removed.

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
...
+* Fluent Bit [1.8.15](https://fluentbit.io/announcements/v1.8.15/)
+* Amazon CloudWatch Logs for Fluent Bit 1.7.0
+* Amazon Kinesis Streams for Fluent Bit 1.9.0
+* Amazon Kinesis Firehose for Fluent Bit 1.6.1
+
+Same as `2.23.2`, this release includes the following fix for AWS customers that we are working on getting accepted upstream:
+* Bug - Resolve IMDSv1 fallback error introduced in 2.21.0 [aws-for-fluent-bit:259](https://github.com/aws/aws-for-fluent-bit/issues/259)
+
+New Tool:
+* [firelens-datajet](https://github.com/aws/firelens-datajet) is a holistic log/metric routing software test tool made to flexibly send test data to Fluent Bit in a repeatable, reliable, and portable manner. Create test configuration JSONs that modularly connect test data generators to data senders, customize data generation rate, send data in parallel over various ingestion routes, mock log router destinations, and in the next version, validate destination results.
```
PettitWesley, line 14: I'd prefer if we not say "in the next version" since folks might confuse this with the aws fluent bit version... and also I think its odd to advertise something that doesn't exist yet

PettitWesley (approved):

CHANGELOG.md
zhonghui12: +1. I'd prefer to call out what we have right now rather than what we will have soon.

matthewfala: changed.

PR #323, opened 2022-03-31
load_tests: add tcp load tests to 4x c plugins

matthewfala:
Signed-off-by: Matthew Fala <falamatt@amazon.com>

*Issue #, if available:*

*Description of changes:*


load_tests/load_test.py
```diff
@@ -110,53 +147,102 @@ def run_ecs_tests():
...
+            )
+    else:
+        # Delete associated s3 bucket objects
+        s3 = boto3.resource('s3')
+        bucket = s3.Bucket(os.environ['S3_BUCKET_NAME'])
+        firehose_objects = bucket.objects.filter(Prefix=f'{OUTPUT_PLUGIN}-test/{PLATFORM}/')
+        firehose_objects.delete()
+        return
 
 def delete_testing_resources():
```
zhonghui12, line 247: I am thinking about if we should keep the logic here as `delete_testing_data` as well:
* cloudwatch codebuild project deletes cw group & ecs cluster
* s3, kinesis, firehose plugins delete each folder generated by the output plugin

load_tests/load_test.py
```diff
@@ -169,11 +255,27 @@ def delete_testing_resources():
...
+        destination_identifier = 'std-' + throughput
+    return destination_identifier
+
+def get_destination_identifier_firelens(throughput, input_logger):
+    return 'std-' + throughput
+
 if sys.argv[1] == 'create_testing_resources':
     create_testing_resources()
 elif sys.argv[1] == 'ECS':
+    delete_testing_data()
```
zhonghui12, line 276: I saw this function in line 157 so I guess it is already called before every run? do we still call it here?

load_tests/logger/tcp_logger/fluent.conf
```diff
@@ -0,0 +1,44 @@
...
+     auto_retry_requests false
+
+[OUTPUT]
+     Name cloudwatch_logs
+     Match cloudwatch
+     log_stream_name ${THROUGHPUT}
+     log_group_name ${CW_LOG_GROUP_NAME}
+     auto_create_group true
+     region us-west-2
+     log_key log
```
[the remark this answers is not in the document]
zhonghui12: I think on fluent bit 1.9, `auto_retry_request` will be default to true. But for now,  we should explicitly keep it set to true in configs

load_tests/load_test.py
```diff
@@ -53,17 +68,38 @@ def calculate_total_input_number(throughput):
...
         '$TASK_EXECUTION_ROLE_ARN': os.environ['LOAD_TEST_TASK_EXECUTION_ROLE_ARN'],
-        '$FLUENT_BIT_IMAGE': os.environ['FLUENT_BIT_IMAGE'],
-        '$APP_IMAGE': os.environ['APP_IMAGE'],
-        '$LOGGER_RUN_TIME_IN_SECOND': str(LOGGER_RUN_TIME_IN_SECOND),
+
+        # Plugin Specific Environment Variables
         'cloudwatch': {'$CW_LOG_GROUP_NAME': os.environ['CW_LOG_GROUP_NAME']},
-        'firehose': {'$DELIVERY_STREAM': os.environ[f'FIREHOSE_TEST_{throughput}']},
-        'kinesis': {'$STREAM': os.environ[f'KINESIS_TEST_{throughput}']},
+        'firehose': {'$DELIVERY_STREAM_PREFIX': f'{PREFIX}{PLATFORM}-firehoseTest-deliveryStream'},
```
zhonghui12, line 101: we can delete this part if you set the env vars like this: https://github.com/aws/aws-for-fluent-bit/blob/mainline/load_tests/setup_test_environment.sh#L12

load_tests/task_definitions/cloudwatch.json
```diff
@@ -2,13 +2,30 @@
...
 	"taskRoleArn": "$TASK_ROLE_ARN",
 	"executionRoleArn": "$TASK_EXECUTION_ROLE_ARN",
+	"networkMode": "bridge",
 	"containerDefinitions": [
 		{
 			"essential": true,
 			"image": "$FLUENT_BIT_IMAGE",
 			"name": "log_router",
+			"environment": [
+				{ "name": "OUTPUT_PLUGIN", "value": "$OUTPUT_PLUGIN" },
```
zhonghui12, line 12: kind of confused here. why we need to set these env vars in the container

zhonghui12: oh is it for config in S3?

zhonghui12 (commented):
Overall looks good. Thanks for the hard work! Have a few comments here

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
```diff
@@ -30,49 +32,52 @@ def __init__(self, scope: core.Construct, construct_id: str, **kwargs) -> None:
...
-                                                                              interval_in_seconds=60,
-                                                                              size_in_m_bs=50
+        for destination in DESTINATION_LIST:
+          for throughput in THROUGHPUT_LIST:
+              caps_identifier = destination.capitalize().replace("-", "") + throughput.capitalize()
+              identifier = destination + throughput
+              # Data streams and related delivery streams for kinesis test
+              names[PLATFORM+'_kinesis_stream_'+identifier] = kinesis.Stream(self, PLATFORM+'KinesisStream'+caps_identifier,
+                                                                            stream_name=PREFIX+PLATFORM+'-kinesisStream-'+identifier,
+                                                                            shard_count=50)
```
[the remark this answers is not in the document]
matthewfala: I ran into some account shard limits, I think you cant have more than 200 shards total? Does that sound incorrect?

load_tests/load_test.py
matthewfala: Discussed separately, Zhonghui explained how the system works and we think that the delete stack will also delete the cloudwatch log group and get rid of the data.

load_tests/load_test.py
matthewfala: I think this is safe to remove! good catch

load_tests/logger/tcp_logger/fluent.conf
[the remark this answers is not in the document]
matthewfala: Will change.

matthewfala (commented):
Thank you for the review!

load_tests/load_test.py
matthewfala: I changed it from stream to stream_prefix, since there are now the following streams

```
# Kinesis
load-test-fluent-bit-ecs-kinesisStream-20m
load-test-fluent-bit-ecs-kinesisStream-25m
load-test-fluent-bit-ecs-kinesisStream-30m

load-test-fluent-bit-ecs-kinesisStream-std-20m
load-test-fluent-bit-ecs-kinesisStream-std-25m
load-test-fluent-bit-ecs-kinesisStream-std-30m

# Firehose
load-test-fluent-bit-ecs-firehoseTest-deliveryStream-20m
load-test-fluent-bit-ecs-firehoseTest-deliveryStream-25m
load-test-fluent-bit-ecs-firehoseTest-deliveryStream-30m

load-test-fluent-bit-ecs-kinesisTest-deliveryStream-std-20m
load-test-fluent-bit-ecs-kinesisTest-deliveryStream-std-25m
```

If we add eks, then we'll need to multiply these streams yet again. to: 
```
# Kinesis
load-test-fluent-bit-eks-kinesisStream-20m
[...]
```

I would like the env vars set on this script to be as minimal as possible. Since we can fully derive the stream prefix from Prefix and Platform, I think that we shouldn't pollute the env vars with combined information, as to make things more complicated. Otherwise it's going to become difficult when we have multiple platforms, prefixes, and postfixes ("" and "-std-")

What do you think?

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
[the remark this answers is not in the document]
zhonghui12: So these streams will not be deleted once we create them, like a one-time setup. Right now our highest throughput is 30mb/s so either 50 or 100 is fine. But our goal is to achieve higher throughput like 75mb/s in the future, and at that time 50 is not enough. It is not a big case for now. Just want to save efforts for future. Also, I don't know about 200 shards limit as I already created three streams in our account and for each is 100 which works? Not sure what you encounter here

load_tests/load_test.py
zhonghui12: I agree. So we can delete the duplicate part in `setup_test_environment.sh`?

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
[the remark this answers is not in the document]
matthewfala: Oh, actually it's 500 in us-west-2 and 200 in some other regions. Either way with 6 streams at 100 shards each, we run out of shards. Unless we request a Quota Increase, which I think is the best option. (increase quota to 1200)

> The default shard quota is 500 shards per AWS account for the following AWS regions: US East (N. Virginia), US West (Oregon), and Europe (Ireland). For all other regions, the default shard quota is 200 shards per AWS account. This limit is only applicable for data streams with the provisioned capacity mode.
>
> To request the shards per data stream quota increase, follow the procedure outlined in [Requesting a Quota Increase](https://docs.aws.amazon.com/servicequotas/latest/userguide/request-quota-increase.html).

See [link](https://docs.aws.amazon.com/streams/latest/dev/service-sizes-and-limits.html)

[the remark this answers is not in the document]
matthewfala: Also, @zhonghui12 Thank you for so carefully reviewing this. I'm gonna change back to 100 and we'll request a shard limit increase.

load_tests/load_test.py
matthewfala: Keeping it simple for now and deleting it from setup_test_environment.sh. As discussed off line, it may be a good idea to have a resource name provider python module that is shared between the creation and load test script. This module could contain 
1. parameter values: throughput, platform, etc
2. parameter->name functions

That way we can abstract the parameter values, such as platforms, from the resources.

Will hold off from this change till a load test improvement PR.

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
```diff
@@ -8,6 +9,7 @@
     core,
 )
 
+DESTINATION_LIST = ["", "std-"] # "" is the destination tag for logs coming from non-stdstream input
```
PettitWesley, line 12: I feel like if this is what it is... "DESTINATION" is the wrong word since this is an input list? Also can it not be an empty string somehow? This does not seem intuitive to understand...

load_tests/load_test.py
PettitWesley: What created the log group? if Fluent Bit created it with `auto_create_group On` (which I see in this PR) then FB created it and CFN delete stack will not delete it. CFN delete stack only deletes things which the stack explicitly created... we should use that, create all log groups in CFN and then set `auto_create_group Off`. This is the best practice actually, and its what CDK does for the awslogs driver for example IIRC, it doesn't use the auto create group option on the log group and instead provisions the log group in the stack, which allows you to delete it later when you delete the stack.

load_tests/load_test.py
```diff
@@ -17,6 +18,20 @@
 else:
     THROUGHPUT_LIST = json.loads(os.environ['THROUGHPUT_LIST'])
 
+# Input Logger Data
+INPUT_LOGGERS = [
+    {
+        "name": "stdstream",
+        "logger_image": "075490442118.dkr.ecr.us-west-2.amazonaws.com/load-test-fluent-bit-app-image:latest",
+        "fluent_config_file_path": "./load_tests/logger/stdout_logger/fluent.conf"
```
PettitWesley, line 26: This file appears to actually be empty... and it looks like we just use the task def to set the output options in all these tests? Why is a config file needed?

PettitWesley: Oh I see, this is because for TCP tests you need config file for the TCP input, but not for the other tests? Would be nicer if you could just put NULL here, but not a big deal.

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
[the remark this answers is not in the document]
zhonghui12: no problem @matthewfala . Actually in this case, I think 80 should be fine. As our highest throughput we've tested is 75mb/s. 80*6 = 480. In this way, we don't need to ask for increase for now.

load_tests/load_test.py
zhonghui12: @PettitWesley we created the log group here: https://github.com/aws/aws-for-fluent-bit/blob/mainline/load_tests/create_testing_resources/app.py#L16. We set the generated log group in task definition and yeah, I guess we should set `auto_create_group Off`

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
zhonghui12: +1, `DESTINATION`sounds more like our output plugins like s3, cloudwatch..

matthewfala: Can we call it DESTINATION_PREFIX_LIST? Because that is what it is. We need a separate destination for TCP and STDOUT because it is possible for error logs to get sent to stdout if the log4j tcp fails. We want to keep these error logs separate from the TCP logs, but still keep them so that we can add some validation in the future if needed. What do you think?

PettitWesley: I still feel that's confusing and that future maintainers of the code might not understand it- it either needs a comment explaining, or why can't it be `LOG_SOURCE_PREFIX_LIST` or `INPUT_PREFIX_LIST` or something like that? This still seems input side, not output side... can you explain how this is related to the output?

matthewfala: Addressed in this new PR into tcp-load-tests https://github.com/fala-aws/aws-for-fluent-bit/pull/2

matthewfala: This should help keep separate of the new, unchecked changes for the Load Test general improvements
- Extract the resource name resolution
- Enhance the markdown output
- Add bar raising logic

load_tests/load_test.py
matthewfala: Added to load test improvement PR

load_tests/load_test.py
matthewfala: To keep the control flow similar for the two input loggers and any other future loggers, let's not change this.

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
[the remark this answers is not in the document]
matthewfala: Still need to do this. Will change after the Improvement PR is merged in: https://github.com/fala-aws/aws-for-fluent-bit/pull/2

load_tests/load_test.py
matthewfala: Still need to do this. Will change after the Improvement PR is merged in: https://github.com/fala-aws/aws-for-fluent-bit/pull/2

load_tests/logger/tcp_logger/fluent.conf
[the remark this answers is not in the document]
matthewfala: done.

load_tests/create_testing_resources/kinesis_s3_firehose/app.py
matthewfala: Merged

matthewfala:
I think that this PR is ready for merge. What do you think @zhonghui12?

zhonghui12:
Based on offline discussion, I think we need to replace the dark dashes with green checks in markdown

buildspec_load_test.yml
```diff
@@ -28,10 +28,16 @@ phases:
...
+      # Will automatically refresh credentials after 1 hour for 5 hours if needed.
+      - |
+        for i in {1..5}
+        do wait
+        CREDS=`aws sts assume-role --role-arn ${LOAD_TEST_CFN_ROLE_ARN} --role-session-name load-test-cfn`
+        export AWS_ACCESS_KEY_ID=`echo $CREDS | jq -r .Credentials.AccessKeyId`
+        export AWS_SECRET_ACCESS_KEY=`echo $CREDS | jq -r .Credentials.SecretAccessKey`
+        export AWS_SESSION_TOKEN=`echo $CREDS | jq -r .Credentials.SessionToken`
+        sleep 3590
+        done &
```
PettitWesley, line 40: won't this block the other code from running? (the code below this) until the while loop finishes? Or does the `do wait` bit do something I don't know about where this is a sub-process?

matthewfala: this was a subprocess. because I added the & at the end of the command. It doesn't work great because python caches the credentials.

Had to remove this system.

matthewfala:
@PettitWesley  & @zhonghui12 
Would you review the following PR into this PR? It resolves the load test credential issues.
https://github.com/fala-aws/aws-for-fluent-bit/pull/8/files

matthewfala:
Would you please review the new changes? @PettitWesley 
https://github.com/fala-aws/aws-for-fluent-bit/pull/9/files

load_tests/load_test.py
```diff
@@ -3,22 +3,49 @@
...
+IS_TASK_DEFINITION_PRINTED = True
 PLATFORM = os.environ['PLATFORM'].lower()
 OUTPUT_PLUGIN = os.environ['OUTPUT_PLUGIN'].lower()
 TESTING_RESOURCES_STACK_NAME = os.environ['TESTING_RESOURCES_STACK_NAME']
 PREFIX = os.environ['PREFIX']
 EKS_CLUSTER_NAME = os.environ['EKS_CLUSTER_NAME']
-LOGGER_RUN_TIME_IN_SECOND = 600
-BUFFER_TIME_IN_SECOND = 300
+LOGGER_RUN_TIME_IN_SECOND = 11
+BUFFER_TIME_IN_SECOND = 70
```
matthewfala, line 19: I'll change this back when we merge

load_tests/load_test.py
```diff
@@ -116,57 +192,236 @@ def create_testing_resources():
...
+    return { x[0]: x[1] for x in list(
+        filter(lambda f: len(f) == 2,
+            map(lambda x: x.split(",  "), validationResultString.decode("utf-8").split("\n"))
+        ))}
+
+def get_validation_output(logger_name, throughput, test_results):
+    return list(filter(lambda r: r["input_logger"]["name"] == logger_name and
+            int(r["test_configuration"]["input_configuration"]["throughput"].replace("m", "")) == throughput, test_results))[0]["parsed_validation_output"]
+
+def format_test_results_to_markdown(test_results):
```
PettitWesley, line 319: yay

PettitWesley (approved):
This is a lot of code.

PR #336, opened 2022-04-25
integ: fix arm64 build

PettitWesley:

In all of our pipeline code, we use `arm64` for 64 bit arm, but in these scripts, we were using the output of `uname` which uses `aarch64`. This fix ensures we always  `arm64` everywhere. 


matthewfala (approved):
LGTM, tested added commands locally. They work.
```
ARCHITECTURE="sam"
if [ "$ARCHITECTURE" = "aarch64" ]; then
    ARCHITECTURE="arm64"
fi
echo $ARCHITECTURE
```
`sam`

```
ARCHITECTURE="aarch64"
if [ "$ARCHITECTURE" = "aarch64" ]; then
    ARCHITECTURE="arm64"
fi
echo $ARCHITECTURE
```
`arm64`

PettitWesley:
I tested this on a graviton instance

PR #343, opened 2022-05-05
Add EKS load tests on CloudWatch

zhonghui12:
*Issue #, if available:*

*Description of changes:*
Add EKS load tests on CloudWatch


zhonghui12:
Put up for review first. Need to be held for merge

load_tests/daemonset/cloudwatch.yaml
```diff
@@ -0,0 +1,248 @@
...
+    [OUTPUT]
+        Name                cloudwatch_logs
+        Match               kube.*
+        region              us-west-2
+        log_group_name      $CW_LOG_GROUP_NAME
+        log_stream_prefix   load-test-fluent-bit-
+        auto_create_group   false
+        log_key             log
+        workers             1
+        auto_retry_requests true
```
PettitWesley, line 91: we should configure a retry_limit on the outputs right?

load_tests/daemonset/cloudwatch.yaml
```diff
@@ -0,0 +1,248 @@
...
+        - name: fluentbitcloudwatch$THROUGHPUT
+          mountPath: /var/fluent-bit/cloudwatch/$THROUGHPUT
+        - name: varlog
+          mountPath: /var/log
+          readOnly: true
+        - name: varlibdockercontainers
+          mountPath: /var/lib/docker/containers
+          readOnly: true
+        - name: cloudwatch-$THROUGHPUT-config
+          mountPath: /fluent-bit/etc/
```
PettitWesley, line 169: why is $THROUGHPUT in these names and paths?

load_tests/daemonset/cloudwatch.yaml
zhonghui12: good catch. I will add it. And also, I am keeping editing this config to improve the result now so it might still be updated based on this version

load_tests/daemonset/cloudwatch.yaml
zhonghui12: It's for this line: https://github.com/aws/aws-for-fluent-bit/blob/d4c55cac1794d04803e40ecea9281af5492a6508/load_tests/daemonset/cloudwatch.yaml#L34 and https://github.com/aws/aws-for-fluent-bit/blob/d4c55cac1794d04803e40ecea9281af5492a6508/load_tests/daemonset/cloudwatch.yaml#L239.  Different throughput tests need different configs so I make it different name (we need different configmap for different tests, otherwise exiting tests will be modified)

load_tests/logger/log_generator.c
```diff
@@ -54,5 +55,9 @@ int main()  {
         msleep(1000+startSeconds-endSeconds);
     }
 
+    // Uncomment below two lines to Keep the generator running for 1 hour
+    // fflush(stdout);
+    // sleep(3600);
```
DrewZhang13, line 60: can we make the running time as a static field or flag to define instead of comment and uncomment in code?

load_tests/load_test.py
```diff
@@ -6,36 +6,34 @@
 import subprocess
 from datetime import datetime, timezone
 
-PLATFORM = os.environ['PLATFORM']
+PLATFORM = os.environ['PLATFORM'].lower()
```
DrewZhang13, line 9: how we decide when the os environment is upper case and when is lower case? could this be unified? Same comment for environment under this line.

load_tests/logger/log_generator.c
zhonghui12: sure

load_tests/load_test.py
zhonghui12: it is set upper case in our pipeline environment just because by convention, environment variables are written in UPPERCASE, but it is used as lower case in our script so we convert it here. Hmm, I am fine to change it but want to discuss with the team first

PettitWesley: I think this makes sense

load_tests/load_test.py
```diff
@@ -6,36 +6,34 @@
...
 OUTPUT_PLUGIN = os.environ['OUTPUT_PLUGIN'].lower()
 TESTING_RESOURCES_STACK_NAME = os.environ['TESTING_RESOURCES_STACK_NAME']
 PREFIX = os.environ['PREFIX']
+EKS_CLUSTER_NAME = os.environ['EKS_CLUSTER_NAME']
 LOGGER_RUN_TIME_IN_SECOND = 600
 BUFFER_TIME_IN_SECOND = 300
+NUM_OF_EKS_NODES = 4
```
PettitWesley, line 16: Can you add a README explaining how the tests work and how tests/pods are distributed between nodes and how someone could run these themselves and etc? Either in this PR or in a separate one.

PettitWesley (approved):

load_tests/daemonset/cloudwatch.yaml
```diff
@@ -0,0 +1,248 @@
...
+    @INCLUDE application-log.conf
+  
+  application-log.conf: |
+    [INPUT]
+        Name                tail
+        Tag                 kube.*
+        Path                /var/log/containers/ds-cloudwatch-$THROUGHPUT*
+        Docker_Mode         On
+        Parser              docker
+        DB                  /var/fluent-bit/cloudwatch/$THROUGHPUT/flb_log.db
```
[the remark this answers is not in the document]
zhonghui12: yeah but we have some other vars need to be replaced (not in fluent bit config) so I just use a script to replace all of them in advance: https://github.com/aws/aws-for-fluent-bit/pull/343/files#diff-3807f668d5ac75597e9a379cb6f8da0e5fb87bff143e946df9071796348e5fb5R172. rather than set env vars in fluent bit

load_tests/load_test.py
PettitWesley: Doesn't this make the value of the env var lower case, Not the env var itself? Also I was saying I agree with Zhonghui, env vars are always upper case I think.

zhonghui12: yeah, not the env var itself

PR #374, opened 2022-06-28
ygloa--init process

Galaoaoa:
*Description of changes:*

Change Makefile: 
Add a "make init" command, which has the same function as original"make release", will also compile init_process.go, and finally build the image "amazon/aws-for-fluent-bit-withinit:latest". 
(Because I use M1 chip to develop, so temporarily added "--platform=linux/amd64" to build the linux/amd64 image, it will be removed in the future, and will add "--no-cache" keyword)

Create Dockerfile.initprocess: 
It is used to compile init_process.go to generate init process binary

Create Dockerfile.fbwithinit: 
This file has the same function as the original "Dockerfile". Compare to the original "Dockerfile", it copies init process binary into the new fluent bit image and It will finally execute "init_process_entrypoint.sh"

Create init_process_entrypoint.sh: 
This file is used to run init process binary and run "invoker.sh". 
"invoke.sh" will execute export command to set ECS Metadata, and finally invoke Fluent Bit  

Create init_process.go:

Add two config files "stdOutput.conf" and "s3Output.conf" in "ecs" folder for testing. Which will be removed in the future.

Note: This is the first submission. Will make changes based on team members' comments and suggestions.


init_process.go
```diff
@@ -0,0 +1,409 @@
...
+	ECS_TASK_ID         string `json:"TaskID"`
+	ECS_FAMILY          string `json:"Family"`
+	ECS_LAUNCHTYPE      string `json:"LaunchType"` //If container agent is under version 1.45.0, this env var will be an empty string
+	ECS_REVISION        string `json:"Revision"`
+	ECS_TASK_DEFINITION string `json:"TaskDefinition"`
+}
+
+//get ECS Task Metadata via endpoint V4
+func getECSTaskMetadata() ECSTaskMetadata {
+	var ecs_task_metadata ECSTaskMetadata
```
zhonghui12, line 71: nit: I think we can just call it `metadata`

init_process.go
```diff
@@ -0,0 +1,409 @@
...
+		fmt.Println(err)
+		log.Fatalf("Cannot find built-in config file: %s\n", path[1:])
+	}
+	defer src.Close()
+
+	dstPath := ConfigFileFolderPath + "/" + fileName[len(fileName)-1]
+	dst, err := os.Create(dstPath)
+	if err != nil {
+		fmt.Println(err)
+		log.Fatalln("Cannot creat config file folder")
```
zhonghui12, line 219: why do we need to create a separate folder for built-in configs? We can just use it as it is included in our repo: /ecs/...

init_process.go
```diff
@@ -0,0 +1,409 @@
...
+
+	io.Copy(dst, src)
+}
+
+//download S3 config file to config file folder
+func getS3ConfigFile(arn string) {
+	if !existS3er {
+		// creat global s3 client
+		// s3er = s3.New(session.Must(session.NewSession()))
+		s3er = s3.New(session.Must(session.NewSession(&aws.Config{
```
zhonghui12, line 231: createS3 Client can be a separate function

init_process.go
```diff
@@ -0,0 +1,409 @@
...
+			//if not specify region here, missingregion error will raise when get bucket location
+			//can be any region
+			Region: aws.String("us-east-2"),
```
zhonghui12, line 234: I think it would be better if the region is the same as `metadata.region`? like we don't specify it. Or it can be if we have ecs metadata, then it is `metadata.region`, Otherwise it can be us-east-1

init_process.go
```diff
@@ -0,0 +1,409 @@
...
+	_, err = mainConfigFile.WriteString(writeContent)
+	if err != nil {
+		fmt.Println(err)
+		log.Fatalf("Cannot write %s in main config file: %s\n", writeContent[:len(writeContent)-2], mainConfigFilePath)
+	}
+}
+
+//change the cammand if needed.
+//add modified command to the init_process_entrypoint.sh
+func changeCommandToAddR(parserFilePath string) {
```
zhonghui12, line 348: maybe we can name it `updateCommand`

zhonghui12 (commented):
Good work! Left some comments. Also

1. please list all cases you have tested
2. for naming, I think we should make it consistent, either use `_` like `ecs_metadata` or lower+upper cases like `ecsMetadata`.
3. I also see some extra spaces/ logger/ print in code, we need to remove them when we do review

Makefile
```diff
@@ -13,6 +13,12 @@
...
 
+.PHONY: init
+init:
+	docker build -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins --platform=linux/amd64  .
+	docker build -t aws-fluent-bit-initprocess:latest -f Dockerfile.initprocess --platform=linux/amd64  .
+	docker build -t amazon/aws-for-fluent-bit-withinit:latest -f Dockerfile.fbwithinit --platform=linux/amd64  .
+
 .PHONY: release
 release:
 	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
```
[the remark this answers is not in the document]
Galaoaoa: Can I do this? After this, the original image which do not contain the init process can no longer be built

Dockerfile.fbwithinit
```diff
@@ -0,0 +1,143 @@
...
+    /cloudwatch/LICENSE \
+    /fluent-bit/licenses/cloudwatch/
+COPY --from=aws-fluent-bit-plugins:latest /kinesis-streams/THIRD-PARTY \
+    /kinesis-streams/LICENSE \
+    /fluent-bit/licenses/kinesis/
+COPY AWS_FOR_FLUENT_BIT_VERSION /AWS_FOR_FLUENT_BIT_VERSION
+ADD ecs /ecs/
+
+COPY entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
```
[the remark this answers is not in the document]
Galaoaoa: Yea, will remove

ecs/s3Output.conf
```diff
@@ -0,0 +1,8 @@
+[OUTPUT]
+    Name                         s3
+    Match                        *
+    bucket                       ygloatest-result-west2
+    region                       ${AWS_REGION}
```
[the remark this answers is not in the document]
Galaoaoa: Yea, will remove

init_process.go
```diff
@@ -0,0 +1,409 @@
+package main
+
+//TODO:
```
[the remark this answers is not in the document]
Galaoaoa: Yea, will remove. Sorry for the notes that were not removed in this PR..

init_process.go
```diff
@@ -0,0 +1,409 @@
...
+const ConfigFileFolderPath = "init_process_fluent-bit_config_files"
+const MainConfigFilePath = "init_process_fluent-bit.conf"
+const OriginalMainConfigFilePath = "/fluent-bit/etc/fluent-bit.conf"
+const InvokerFilePath = "invoker.sh"
+
+//default command
+var FluentBitCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
+
+//global s3 client
+var s3er *s3.S3
```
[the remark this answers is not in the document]
Galaoaoa: Ok, will change the name

init_process.go
```diff
@@ -0,0 +1,409 @@
...
+//set error log format
+func setErrorLog() {
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	log.SetPrefix("FluentBit Init Process-> ")
+}
+
+//create a invoker.sh
+//which will declare ECS Task Metadata as environment variables
+//and finally invoke Fluent Bit
+func createInvokerFile(filePath string) {
```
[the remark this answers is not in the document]
Galaoaoa: I will adjust and merge these functions

init_process.go
Galaoaoa: This actually came from a discussion with Bingfeng, he suggested that we eventually output all the configuration files used for the customer, and I'm still thinking about how to do it here..

The `cat` command in the [init_process_entrypoint.sh] is also for the same purpose

Will design a more reasonable way and modify it.

init_process.go
Galaoaoa: Yeah, will try to modify

init_process.go
Galaoaoa: Yeah make sense, will modify

init_process.go
Galaoaoa: Yea, will modify. And thank you for all the comments mentioned above :)

PR #377, opened 2022-07-01
Init Process for AWS for Fluent Bit on ECS

Galaoaoa:
### Description 

By submitting this pull request, users can build the new image  `amazon/aws-for-fluent-bit-init`  by using the command  `make init`

### Test 

I have tested the expected functionality of the init process.

Based on the [cloudwatchlogs example](https://github.com/aws-samples/amazon-ecs-firelens-examples/blob/mainline/examples/fluent-bit/cloudwatchlogs/task-definition.json), I added five environment variables in the ECS Task definition:

`aws_fluent_bit_s3_1`		arn:aws:s3:::ygloatest-east1/dummyParser.conf

```
[PARSER]
    Name dummy_test
    Format regex
    Regex ^(?<INT>[^ ]+) (?<FLOAT>[^ ]+) (?<BOOL>[^ ]+) (?<STRING>.+)$
```

`aws_fluent_bit_s3_2`		arn:aws:s3:::ygloatest-east2/dummyFilter.conf

```
[FILTER]
    Name parser
    Match dummy.*
    Key_Name data
    Parser dummy_test
```

`aws_fluent_bit_s3_3`		arn:aws:s3:::ygloatest-west2/dummyInput.conf

```
[INPUT]
    Name dummy
    Tag  dummy.data
    Dummy {"data":"100 0.5 true This is example"}
```

`aws_fluent_bit_file_1`		/ecs/s3Output.conf

```
[OUTPUT]
    Name                         s3
    Match                        *
    bucket                       ygloatest-result-west2
    region                       ${AWS_REGION}
    total_file_size              1M
    upload_timeout               1m
    use_put_object               On
```

`aws_fluent_bit_file_2`		/ecs/stdOutput.conf

```
[OUTPUT]
    Name stdout
    Match *
```

Ran this task using new image which include init process. The result was as expected. 

The ecs task metadata was successfully set as environment variables and cloud be used in the config file. The main configuration file used @INCLUDE to add the original config file and all config files user specified. 

```
@INCLUDE /fluent-bit/etc/fluent-bit.conf
@INCLUDE /ecs/s3Output.conf
@INCLUDE /ecs/stdOutput.conf
@INCLUDE /init_process_fluent-bit_s3_config_files/dummyFilter.conf
@INCLUDE /init_process_fluent-bit_s3_config_files/dummyInput.conf
```

Fluent Bit command was updated successfully: `/fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so -c /init_process_fluent-bit.conf -R /init_process_fluent-bit_s3_config_files/dummyParser.conf`. 

Finally the logs are forwarded to cloudwatch, S3, and the container stdout.

### Details 

* Change `Makefile`
  Add the `make init` command, which has the same function as original `make release`, will also compile init_process.go, and finally build the image "amazon/aws-for-fluent-bit-init:latest"

* Create `Dockerfile.build_init_process`
  It is used to compile init_process.go to generate init process binary

* Create `Dockerfile.init`
  This file has the same function as the original `Dockerfile`. Compare to the original `Dockerfile`, it copies init process binary into the new fluent bit image and It will finally execute `init_process_entrypoint.sh`

* Create `init_process_entrypoint.sh`
  This file is used to run init process binary and run `invoker.sh`, which will execute export command to set ECS Metadata, and finally invoke Fluent Bit

* Create `init_process.go`

***Note:*** Will redesign the Dockerfile to merge `Dockerfile.build_init_process` in `Dockerfile.init`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,12 @@
 
 all: release
 
+.PHONY: init
+init:
+	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
+	docker build --no-cache -t aws-fluent-bit-initprocess:latest -f Dockerfile.build_init_process .
+	docker build --no-cache -t amazon/aws-for-fluent-bit-init:latest -f Dockerfile.init .
+
 .PHONY: release
 release:
 	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
--- a/init_process.go
+++ b/init_process.go
@@ -1,0 +1,406 @@
+package main
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"reflect"
+	"regexp"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+)
+
+// static file path
+const ConfigFileFolderPath = "init_process_fluent-bit_s3_config_files"
+const MainConfigFilePath = "init_process_fluent-bit.conf"
+const OriginalMainConfigFilePath = "/fluent-bit/etc/fluent-bit.conf"
+const InvokerFilePath = "invoker.sh"
+
+// default Fluent Bit command
+var FluentBitCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
+
+// global s3 client and flag
+var s3Client *s3.S3
+var exists3Client bool = false
+
+// global ecs metadata region
+var metadataReigon string = ""
+
+// set error log format
+func setErrorLog() {
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	log.SetPrefix("FluentBit Init Process -> ")
+}
+
+// create the invoker.sh
+// which will declare ECS Task Metadata as environment variables
+// and finally invoke Fluent Bit
+func createInvokerFile(filePath string) {
+	invokerFile, err := os.Create(filePath)
+	if err != nil {
+		log.Println(err)
+		log.Fatalln("Cannot create fluent bit invoker file")
+	}
+	defer invokerFile.Close()
+}
+
+type ECSTaskMetadata struct {
+	AWS_REGION          string `json:"AWSRegion"`
+	ECS_CLUSTER         string `json:"Cluster"`
+	ECS_TASK_ARN        string `json:"TaskARN"`
+	ECS_TASK_ID         string `json:"TaskID"`
+	ECS_FAMILY          string `json:"Family"`
+	ECS_LAUNCHTYPE      string `json:"LaunchType"` //If container agent is under version 1.45.0, this env var will be an empty string
+	ECS_REVISION        string `json:"Revision"`
+	ECS_TASK_DEFINITION string `json:"TaskDefinition"`
+}
+
+// get ECS Task Metadata via endpoint V4
+func getECSTaskMetadata() ECSTaskMetadata {
+	var metadata ECSTaskMetadata
+
+	ecsTaskMetadataEndpointV4 := os.Getenv("ECS_CONTAINER_METADATA_URI_V4")
+	if ecsTaskMetadataEndpointV4 == "" {
+		log.Panicln("Unable to get ECS Metadata")
+	}
+
+	res, err := http.Get(ecsTaskMetadataEndpointV4 + "/task")
+	if err != nil {
+		log.Panicln("failed to get ECS Metadata via HTTP Get")
+	}
+
+	response, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		log.Panicln("failed to read response")
+	}
+	res.Body.Close()
+
+	err = json.Unmarshal(response, &metadata)
+	if err != nil {
+		log.Println(err)
+		log.Panicln("failed to unmarshal ECS metadata")
+	}
+
+	ARN, err := arn.Parse(metadata.ECS_TASK_ARN)
+	if err != nil {
+		log.Println(err)
+		log.Panicln("failed to parse ECS TaskARN")
+	}
+
+	resourceID := strings.Split(ARN.Resource, "/")
+	taskID := resourceID[len(resourceID)-1]
+	metadata.ECS_TASK_ID = taskID
+	metadata.AWS_REGION = ARN.Region
+	metadata.ECS_TASK_DEFINITION = metadata.ECS_FAMILY + ":" + metadata.ECS_REVISION
+
+	return metadata
+}
+
+// set ECS Task Metadata as environment variables in the invoker.sh
+func setECSTaskMetadataAsEnvVar(metadata ECSTaskMetadata, filePath string) {
+	t := reflect.TypeOf(metadata)
+	v := reflect.ValueOf(metadata)
+
+	initProcessEntrypointFile, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY, 0777)
+	if err != nil {
+		log.Println(err)
+		log.Fatalf("Unable to read init process entrypoint file: %s\n", filePath)
+	}
+	defer initProcessEntrypointFile.Close()
+
+	for i := 0; i < t.NumField(); i++ {
+		if v.Field(i).Interface().(string) == "" {
+			continue
+		}
+		writeContent := "export " + t.Field(i).Name + "=" + v.Field(i).Interface().(string) + "\n"
+		_, err = initProcessEntrypointFile.WriteString(writeContent)
+		if err != nil {
+			log.Println(err)
+			log.Fatalf("Cannot write %s in init process entrypoint file\n", writeContent[:len(writeContent)-2])
+		}
+	}
+}
+
+// create a new main config file
+func createMainConfigFile(filePath string) {
+	mainConfigFile, err := os.Create(filePath)
+	if err != nil {
+		log.Println(err)
+		log.Fatalln("Cannot create main config file")
+	}
+	defer mainConfigFile.Close()
+}
+
+// add @INCLUDE original main config file to the new main config file
+func includeOriginalMainConfigFile(mainConfigFilePath, originalMainConfigFilePath string) {
+	mainConfigFile, err := os.OpenFile(mainConfigFilePath, os.O_APPEND|os.O_WRONLY, 0777)
+	if err != nil {
+		log.Println(err)
+		log.Fatalf("Unable to read main config file: %s\n", mainConfigFilePath)
+	}
+	defer mainConfigFile.Close()
+
+	writeContent := "@INCLUDE " + originalMainConfigFilePath + "\n"
+	_, err = mainConfigFile.WriteString(writeContent)
+	if err != nil {
+		log.Println(err)
+		log.Fatalf("Cannot write %s in main config file: %s\n", writeContent[:len(writeContent)-2], mainConfigFilePath)
+	}
+}
+
+// create Fluent Bit command to use new main config file
+func createCommand(command *string, filePath string) {
+	*command = *command + " -c /" + filePath
+}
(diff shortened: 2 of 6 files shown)


zhonghui12:
Actually I don't understand that why do you need to create new files like `/ecs/s3Output.conf`? When we say built-in, we mean the templates here:https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs  and these files will be build into fluent bit container: https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile#L128. Built-in files are not custom files, they are just the templates provided by us. I am not saying you are wrong, if you create & copy more files into /ecs/ path in the container and it is also okay and should work. I am just saying you could just reuse our existing ones.

Makefile
```diff
@@ -13,6 +13,12 @@
...
 
+.PHONY: init
+init:
+	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
+	docker build --no-cache -t aws-fluent-bit-initprocess:latest -f Dockerfile.build_init_process .
+	docker build --no-cache -t amazon/aws-for-fluent-bit-init:latest -f Dockerfile.init .
+
 .PHONY: release
 release:
 	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
```
zhonghui12, line 24: add init build in `make release` as well

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+)
+
+// static file path
+const ConfigFileFolderPath = "init_process_fluent-bit_s3_config_files"
```
zhonghui12, line 21: Let's don't include both - and _. and make these starting with lower cases. I prefer:
```
const s3FilePath = "fluent-bit-init-s3-files"
const mainConfigFilePath = "fluent-bit-init.conf"
const originalMainConfigPath = "/fluent-bit/etc/fluent-bit.conf"
const invokerFilePath = "invoker.sh"
```

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+const MainConfigFilePath = "init_process_fluent-bit.conf"
+const OriginalMainConfigFilePath = "/fluent-bit/etc/fluent-bit.conf"
+const InvokerFilePath = "invoker.sh"
+
+// default Fluent Bit command
+var FluentBitCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
```
zhonghui12, line 27: `fluentBitCommand`

zhonghui12: just for consistency: https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit/blob/mainline/cloudwatch/cloudwatch_test.go

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+// set error log format
+func setErrorLog() {
+	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
+	log.SetPrefix("FluentBit Init Process -> ")
+}
+
+// create the invoker.sh
+// which will declare ECS Task Metadata as environment variables
+// and finally invoke Fluent Bit
+func createInvokerFile(filePath string) {
```
zhonghui12, line 45: I cannot see any difference for this function and `createMainConfigFile` in line 132. Both accept a file path and then create a file. Why do you make them separately?

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+	defer invokerFile.Close()
+}
+
+type ECSTaskMetadata struct {
+	AWS_REGION          string `json:"AWSRegion"`
+	ECS_CLUSTER         string `json:"Cluster"`
+	ECS_TASK_ARN        string `json:"TaskARN"`
+	ECS_TASK_ID         string `json:"TaskID"`
+	ECS_FAMILY          string `json:"Family"`
+	ECS_LAUNCHTYPE      string `json:"LaunchType"` //If container agent is under version 1.45.0, this env var will be an empty string
```
zhonghui12, line 60: `ECS_LAUNCH_TYPE`

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	res.Body.Close()
+
+	err = json.Unmarshal(response, &metadata)
+	if err != nil {
+		log.Println(err)
+		log.Panicln("failed to unmarshal ECS metadata")
+	}
+
+	ARN, err := arn.Parse(metadata.ECS_TASK_ARN)
```
zhonghui12, line 91: `arn`

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+
+// global s3 client and flag
+var s3Client *s3.S3
+var exists3Client bool = false
+
+// global ecs metadata region
+var metadataReigon string = ""
+
+// set error log format
+func setErrorLog() {
```
zhonghui12, line 37: for error log, let's do the same thing as our go plugins: https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit/blob/mainline/cloudwatch/cloudwatch.go#L178-L200. If return, print the error. Or use `logrus` to log

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+
+type ECSTaskMetadata struct {
+	AWS_REGION          string `json:"AWSRegion"`
+	ECS_CLUSTER         string `json:"Cluster"`
+	ECS_TASK_ARN        string `json:"TaskARN"`
+	ECS_TASK_ID         string `json:"TaskID"`
+	ECS_FAMILY          string `json:"Family"`
+	ECS_LAUNCHTYPE      string `json:"LaunchType"` //If container agent is under version 1.45.0, this env var will be an empty string
+	ECS_REVISION        string `json:"Revision"`
+	ECS_TASK_DEFINITION string `json:"TaskDefinition"`
```
zhonghui12, line 62: you can add a comment here for what it means

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+	taskID := resourceID[len(resourceID)-1]
+	metadata.ECS_TASK_ID = taskID
+	metadata.AWS_REGION = ARN.Region
+	metadata.ECS_TASK_DEFINITION = metadata.ECS_FAMILY + ":" + metadata.ECS_REVISION
+
+	return metadata
+}
+
+// set ECS Task Metadata as environment variables in the invoker.sh
+func setECSTaskMetadataAsEnvVar(metadata ECSTaskMetadata, filePath string) {
```
zhonghui12, line 107: nit: I think  `setECSTaskMetadata` is enought

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+// create Fluent Bit command to use new main config file
+func createCommand(command *string, filePath string) {
+	*command = *command + " -c /" + filePath
+}
+
+// create a folder to store S3 config files user specified
+func createS3ConfigFileFolder(folderPath string) {
+	os.Mkdir(folderPath, os.ModePerm)
```
zhonghui12, line 165: I am wondering if there is a way to create a function to create either a file or a path based on the string? I think it should..

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+		log.Println(err)
+		log.Fatalln("Cannot creat a new session")
+	}
+
+	// need to specify session region!
+	s3Downloader := s3manager.NewDownloader(sess)
+
+	// download file and store
+	s3FileName := strings.SplitN(s3FilePath, "/", -1)
+	fileFromS3, err := os.Create(ConfigFileFolderPath + "/" + s3FileName[len(s3FileName)-1])
```
zhonghui12, line 267: Is it the same as `createfile` function?

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+		log.Fatalf("Unable to read config files in %s folder\n", folderPath)
+	}
+
+	for _, file := range fileInfos {
+		content_b, err := ioutil.ReadFile(folderPath + "/" + file.Name())
+		if err != nil {
+			log.Println(err)
+			log.Fatalf("Cannot open file: %s\n", folderPath+"/"+file.Name())
+		}
+		content := string(content_b)
```
zhonghui12, line 313: this part is the same as `processBuiltInConfigs`.. we should put all the same thing into one function to execute. we don't split them by names but by the thing they will do. either it is built-in files or s3 files, we can use one same function to process them

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+			updateCommand("/" + folderPath + "/" + file.Name())
+		} else {
+			// this is not a parser config file. @INCLUDE
+			writeInclude("/"+folderPath+"/"+file.Name(), MainConfigFilePath)
+		}
+	}
+}
+
+// add @INCLUDE in main config file to include all customer specified config files
+func writeInclude(configFilePath, mainConfigFilePath string) {
```
zhonghui12, line 325: Why do we need `includeOriginalMainConfigFile` along? both do include thing

init_process.go
```diff
@@ -0,0 +1,406 @@
...
+func updateCommand(parserFilePath string) {
+	FluentBitCommand = FluentBitCommand + " -R " + parserFilePath
+	log.Println("Command is change to -> " + FluentBitCommand)
+}
+
+// create a init_process_entrypoint.sh
+// which will declare ECS Task Metadata as environment variables
+// and finally invoke Fluent Bit
+func modifyInvokerFile(filePath string) {
+	invokerFile, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY, 0777)
```
zhonghui12, line 352: I see many openfiles in the code. We can have one separate function for it

zhonghui12 (commented):
Please also test more cases like no env input or wrong env input

Galaoaoa:
>

Galaoaoa:
> Actually I don't understand that why do you need to create new files like `/ecs/s3Output.conf`? When we say built-in, we mean the templates here:https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs and these files will be build into fluent bit container: https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile#L128. Built-in files are not custom files, they are just the templates provided by us. I am not saying you are wrong, if you create & copy more files into /ecs/ path in the container and it is also okay and should work. I am just saying you could just reuse our existing ones.

Yeah, sure I know that, I just want to test the function which to get config files from out built-in folder. And see the result more directly. Cause I can't find a suitable built-in config file in `ecs` folder for my test example so I create a new one just for test this time.

Makefile
Galaoaoa: > 

Yeah sure. For the case no env input, this image can run normally like the original image I have already tested. And for wrong env input, the process will be stoped.
And will test more cases and write them in next PR description. Thank you for this advice!

Galaoaoa: Add init build in `make release`, I will try this after figure out the CMD command issue

init_process.go
Galaoaoa: Already modified

init_process.go
Galaoaoa: Already using the consistent style

init_process.go
Galaoaoa: Thank you for your advice, have modified

init_process.go
Galaoaoa: Ok, have changed to use `logrus `

init_process.go
Galaoaoa: Yeah, already added

init_process.go
Galaoaoa: If it's only for the init process, I can achieve this by determining if there is a` . ` in the file path. But I don't think that's very safe..  So I ended up setting create file and create folder as two functions

init_process.go
Galaoaoa: Yeah, already modified

init_process.go
Galaoaoa: Yeah, already modified.

init_process.go
Galaoaoa: Make sense.. already modified

init_process.go
Galaoaoa: Already modified. 
After making these changes, encapsulating duplicate code. I was able to better understand how to write modular code, thank you!

PR #379, opened 2022-07-08
Init Process for AWS for Fluent Bit on ECS

Galaoaoa:
#### Description 

Users can build the new image  `amazon/aws-for-fluent-bit-init:latest` and original image `amazon/aws-for-fluent-bit:latest` as the same time by using the command  `make release`

---

#### Changes from last PR

Code Style

* Change the original code style to use a consistent code style and naming style with the team
* Change code structure for modularization and code reuse

Unit Test

* Add code for unit test in `fluent_bit_init_process_test.go`

Dockerfile

* Merge two dockerfiles into `Dockerfile.init`

---

#### Details of comparison with `aws/aws-for-fluent-bit @ mainline`

* Change `Makefile`
  Change the `make release` command, add `docker build -t amazon/aws-for-fluent-bit-init:latest -f Dockerfile.init .` After executing the `make release` command, will also build the image `amazon/aws-for-fluent-bit-init:latest`

* Create `Dockerfile.init`
  This Dockerfile will build the new image `amazon/aws-for-fluent-bit-init:latest` which include init process and finally execute `fluent_bit_init_entrypoint.sh`

* Create `init` folder
  This folder is used to store files related to the init process (following three files)

* Create `fluent_bit_init_entrypoint.sh`
  This file is used to run init process binary and run `fluent-bit-invoker.sh`, which will execute export command to set ECS Metadata, and finally invoke Fluent Bit

* Create `fluent_bit_init_process.go`

* Create `fluent_bit_init_process_test.go`
  This file is the code of Unit Test for `fluent_bit_init_process.go`

---

#### Tests already performed

**Run image on ECS**

* Without any user specified environment variables. Same effect as using the original image

* Add a Dummy Input config file via setting environment variables. Main config file is changed as expected

* Add a Dummy Input config file, a Parser config file and a related Filter config file via setting environment variables. Main config file is changed as expected, the Fluent Bit Command is changed as expected

* Add  a Dummy Input config file, a Parser config file, a related Filter config file, a S3 Output config file, a stdout Output config file via setting environment variables. Main config file is changed as expected, the Fluent Bit Command is changed as expected. Can receive the log in the container and S3

**Run image locally**

* Without ECS Task Metadata. Can invoke Fluent Bit and print out a warning `[FluentBit Init Process] Unable to get ECS Metadata, ignore this warning if not running on ECS`

### DIFF
--- a/Dockerfile.init
+++ b/Dockerfile.init
@@ -1,0 +1,30 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:latest as init-builder
+
+RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+RUN chmod +x /bin/gimme
+RUN yum upgrade -y && yum install -y tar gzip git
+ENV HOME /home
+RUN /bin/gimme 1.17.9
+ENV PATH ${PATH}:/home/.gimme/versions/go1.17.9.linux.arm64/bin:/home/.gimme/versions/go1.17.9.linux.amd64/bin
+RUN go version
+ENV GO111MODULE on
+RUN go env -w GOPROXY=direct
+
+# Build init process for Fluent Bit
+COPY /init/fluent_bit_init_process.go /
+COPY /go.mod /
+COPY /go.sum /
+RUN go mod tidy \
+    && go build fluent_bit_init_process.go
+
+FROM amazon/aws-for-fluent-bit:latest
+
+RUN mkdir -p /init
+
+COPY --from=init-builder /fluent_bit_init_process /init/fluent_bit_init_process
+
+COPY init/fluent_bit_init_entrypoint.sh /init/fluent_bit_init_entrypoint.sh
+RUN chmod +x /init/fluent_bit_init_entrypoint.sh
+
+# Only last CMD command will be executed, automatically replaces the original entrypoint
+CMD /init/fluent_bit_init_entrypoint.sh
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ all: release
 release:
 	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
 	docker build -t amazon/aws-for-fluent-bit:latest -f Dockerfile .
+	docker build -t amazon/aws-for-fluent-bit:init-latest -f Dockerfile.init .
 
 .PHONY: debug
 debug:
--- a/init/fluent_bit_init_entrypoint.sh
+++ b/init/fluent_bit_init_entrypoint.sh
@@ -1,0 +1,2 @@
+./init/fluent_bit_init_process
+source /init/invoke_fluent_bit.sh
--- a/init/fluent_bit_init_process.go
+++ b/init/fluent_bit_init_process.go
@@ -1,0 +1,388 @@
+package main
+
+import (
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"reflect"
+	"regexp"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/sirupsen/logrus"
+)
+
+// static paths
+const (
+	s3FileDirectoryPath    = "/init/fluent-bit-init-s3-files/"
+	mainConfigFile         = "/init/fluent-bit-init.conf"
+	originalMainConfigFile = "/fluent-bit/etc/fluent-bit.conf"
+	invokeFile             = "/init/invoke_fluent_bit.sh"
+)
+
+var (
+	// default Fluent Bit command
+	baseCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
+
+	// global s3 client and flag
+	s3Client        *s3.S3
+	s3ClientCreated bool = false
+
+	// global ecs metadata region
+	metadataRegion string = ""
+)
+
+// HTTPClient interface
+type HTTPClient interface {
+	Get(url string) (*http.Response, error)
+}
+
+// S3Downloader interface
+type S3Downloader interface {
+	Download(w io.WriterAt, input *s3.GetObjectInput, options ...func(*s3manager.Downloader)) (int64, error)
+}
+
+// all values in the structure are empty strings by default
+type ECSTaskMetadata struct {
+	AWS_REGION          string `json:"AWSRegion"`
+	ECS_CLUSTER         string `json:"Cluster"` // Cluster name
+	ECS_TASK_ARN        string `json:"TaskARN"`
+	ECS_TASK_ID         string `json:"TaskID"`
+	ECS_FAMILY          string `json:"Family"`
+	ECS_LAUNCH_TYPE     string `json:"LaunchType"`     // Task launch type will be an empty string if container agent is under version 1.45.0
+	ECS_REVISION        string `json:"Revision"`       // Revision number
+	ECS_TASK_DEFINITION string `json:"TaskDefinition"` // TaskDefinition = "family:revision"
+}
+
+// get ECS Task Metadata via endpoint V4
+func getECSTaskMetadata(httpClient HTTPClient) ECSTaskMetadata {
+	var metadata ECSTaskMetadata
+
+	ecsTaskMetadataEndpointV4 := os.Getenv("ECS_CONTAINER_METADATA_URI_V4")
+	if ecsTaskMetadataEndpointV4 == "" {
+		logrus.Warnln("[FluentBit Init Process] Unable to get ECS Metadata, ignore this warning if not running on ECS")
+		return metadata
+	}
+
+	res, err := httpClient.Get(ecsTaskMetadataEndpointV4 + "/task")
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Failed to get ECS Metadata via HTTP Get: %s\n", err)
+	}
+
+	response, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Failed to read ECS Metadata from HTTP response: %s\n", err)
+	}
+	res.Body.Close()
+
+	err = json.Unmarshal(response, &metadata)
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Failed to unmarshal ECS metadata: %s\n", err)
+	}
+
+	arn, err := arn.Parse(metadata.ECS_TASK_ARN)
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Failed to parse ECS TaskARN: %s\n", err)
+	}
+
+	resourceID := strings.Split(arn.Resource, "/")
+	taskID := resourceID[len(resourceID)-1]
+	metadata.ECS_TASK_ID = taskID
+	metadata.AWS_REGION = arn.Region
+	metadata.ECS_TASK_DEFINITION = metadata.ECS_FAMILY + ":" + metadata.ECS_REVISION
+
+	// set global ecs metadata region for S3 client
+	metadataRegion = reflect.ValueOf(metadata).Field(0).Interface().(string)
+
+	return metadata
+}
+
+// set ECS Task Metadata as environment variables in the invoke_fluent_bit.sh
+func setECSTaskMetadata(metadata ECSTaskMetadata, filePath string) {
+	invokeFile := openFile(filePath)
+	defer invokeFile.Close()
+
+	// set the FLB_AWS_USER_AGENT env var as "init" to get the image usage
+	initUsage := "export FLB_AWS_USER_AGENT=init\n"
+	_, err := invokeFile.WriteString(initUsage)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Warnf("[FluentBit Init Process] Cannot write %s in the invoke_fluent_bit.sh\n", initUsage[:len(initUsage)-2])
+	}
+
+	t := reflect.TypeOf(metadata)
+	v := reflect.ValueOf(metadata)
+
+	for i := 0; i < t.NumField(); i++ {
+		if v.Field(i).Interface().(string) == "" {
+			continue
+		}
+		writeContent := "export " + t.Field(i).Name + "=" + v.Field(i).Interface().(string) + "\n"
+		_, err := invokeFile.WriteString(writeContent)
+		if err != nil {
+			logrus.Errorln(err)
+			logrus.Fatalf("[FluentBit Init Process] Cannot write %s in the invoke_fluent_bit.sh\n", writeContent[:len(writeContent)-2])
+		}
+	}
+}
+
+// create Fluent Bit command to use "-c" to specify the new main config file
+func createCommand(command *string, filePath string) {
+	*command = *command + " -c " + filePath
+}
+
+// get our built in config files or files from s3
+// process built-in config files directly
+// add S3 config files to directory "/init/fluent-bit-init-s3-files/"
+func getAllConfigFiles() {
+	// get all env vars in the container
+	envs := os.Environ()
+
+	// find all env vars match specified prefix
+	for _, env := range envs {
+		var envKey string
+		var envValue string
+		env_kv := strings.SplitN(env, "=", 2)
+		if len(env_kv) != 2 {
+			logrus.Fatalf("[FluentBit Init Process] Unrecognizable environment variables: %s\n", env)
+		}
+
+		envKey = string(env_kv[0])
+		envValue = string(env_kv[1])
+
+		s3_regex, _ := regexp.Compile("aws_fluent_bit_init_[sS]3")
+		file_regex, _ := regexp.Compile("aws_fluent_bit_init_[fF]ile")
+
+		matched_s3 := s3_regex.MatchString(envKey)
+		matched_file := file_regex.MatchString(envKey)
+
+		// if this env var's value is an arn, download the config file first, then process it
+		if matched_s3 {
+			s3FilePath := getS3ConfigFile(envValue)
+			s3FileName := strings.SplitN(s3FilePath, "/", -1)
+			processConfigFile(s3FileDirectoryPath + s3FileName[len(s3FileName)-1])
+		}
+		// if this env var's value is a path of our built-in config file, process is derectly
+		if matched_file {
+			processConfigFile(envValue)
+		}
+	}
+}
+
+func processConfigFile(path string) {
+	contentBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot open file: %s\n", path)
+	}
+
+	content := string(contentBytes)
+
+	if strings.Contains(content, "[PARSER]") {
+		// this is a parser config file, change command
+		updateCommand(path)
+	} else {
+		// this is not a parser config file. @INCLUDE
+		writeInclude(path, mainConfigFile)
+	}
+}
+
+func getS3ConfigFile(arn string) string {
+	// Preparation for downloading S3 config files
+	if !s3ClientCreated {
+		createS3Client()
+	}
+
+	// e.g. "arn:aws:s3:::user-bucket/s3_parser.conf"
+	arnBucketFile := arn[13:]
+	bucketAndFile := strings.SplitN(arnBucketFile, "/", 2)
+	if len(bucketAndFile) != 2 {
+		logrus.Fatalf("[FluentBit Init Process] Unrecognizable arn: %s\n", arn)
+	}
+
+	bucketName := bucketAndFile[0]
+	s3FilePath := bucketAndFile[1]
+
+	// get bucket region
+	input := &s3.GetBucketLocationInput{
+		Bucket: aws.String(bucketName),
+	}
+
+	output, err := s3Client.GetBucketLocation(input)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot get bucket region of %s + %s, you must be the bucket owner to implement this operation\n", bucketName, s3FilePath)
+	}
+
+	bucketRegion := aws.StringValue(output.LocationConstraint)
+	// Buckets in Region us-east-1 have a LocationConstraint of null
+	// https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#GetBucketLocationOutput
+	if bucketRegion == "" {
+		bucketRegion = "us-east-1"
+	}
+
+	// create a downloader
+	s3Downloader := createS3Downloader(bucketRegion)
+
+	// download file from S3 and store in the directory "/init/fluent-bit-init-s3-files/"
+	downloadS3ConfigFile(s3Downloader, s3FilePath, bucketName, s3FileDirectoryPath)
+
+	return s3FilePath
+}
+
+// create a S3 client as the global S3 client for reuse
+func createS3Client() {
+	region := "us-east-1"
+	if metadataRegion != "" {
+		region = metadataRegion
+	}
+
+	s3Client = s3.New(session.Must(session.NewSession(&aws.Config{
+		// if not specify region here, missingregion error will raise when get bucket location
+		Region: aws.String(region),
+	})))
+
+	s3ClientCreated = true
+}
+
+func createS3Downloader(bucketRegion string) S3Downloader {
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(bucketRegion)},
+	)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalln("[FluentBit Init Process] Cannot creat a new session")
+	}
+
+	// need to specify session region!
+	s3Downloader := s3manager.NewDownloader(sess)
+	return s3Downloader
+}
+
+func downloadS3ConfigFile(s3Downloader S3Downloader, s3FilePath, bucketName, s3FileDirectory string) {
+	s3FileName := strings.SplitN(s3FilePath, "/", -1)
+	fileFromS3 := createFile(s3FileDirectory+s3FileName[len(s3FileName)-1], false)
+	defer fileFromS3.Close()
+
+	_, err := s3Downloader.Download(fileFromS3,
+		&s3.GetObjectInput{
+			Bucket: aws.String(bucketName),
+			Key:    aws.String(s3FilePath),
+		})
+	if err != nil {
+		logrus.Warnf("[FluentBit Init Process] Cannot download %s from s3, retrying...\n", s3FileName)
+
+		_, error := s3Downloader.Download(fileFromS3,
+			&s3.GetObjectInput{
+				Bucket: aws.String(bucketName),
+				Key:    aws.String(s3FilePath),
+			})
+		if error != nil {
+			logrus.Errorln(error)
+			logrus.Fatalf("[FluentBit Init Process] Cannot download %s from s3\n", s3FileName)
+		}
+	}
+}
+
+// use @INCLUDE to add config files to the main config file
+func writeInclude(configFilePath, mainConfigFilePath string) {
+	mainConfigFile := openFile(mainConfigFilePath)
+	defer mainConfigFile.Close()
+
+	writeContent := "@INCLUDE " + configFilePath + "\n"
+	_, err := mainConfigFile.WriteString(writeContent)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot write %s in main config file: %s\n", writeContent[:len(writeContent)-2], mainConfigFilePath)
+	}
+}
+
+// change the fluent bit cammand to use "-R" to specift Parser config file
+func updateCommand(parserFilePath string) {
+	baseCommand = baseCommand + " -R " + parserFilePath
+	logrus.Infoln("[FluentBit Init Process] Command is change to -> " + baseCommand)
+}
+
+// change the invoke_fluent_bit.sh
+// which will declare ECS Task Metadata as environment variables
+// and finally invoke Fluent Bit
+func modifyInvokeFile(filePath string) {
+	invokeFile := openFile(filePath)
+	defer invokeFile.Close()
+
+	_, err := invokeFile.WriteString(baseCommand)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot write %s in invoke_fluent_bit.sh\n", baseCommand)
+	}
+}
+
+// create a file, when flag is true, the file will be closed automatically after creation
+func createFile(filePath string, AutoClose bool) *os.File {
+	if err := os.MkdirAll(filepath.Dir(filePath), 0700); err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot create the Directory: %s\n", filepath.Dir(filePath))
+	}
+
+	file, err := os.Create(filePath)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot create the file: %s\n", filePath)
+	}
+
+	if AutoClose {
+		defer file.Close()
+	}
+
+	return file
+}
+
+func openFile(filePath string) *os.File {
+	file, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY, 0700)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Unable to read %s\n", filePath)
+	}
+	return file
+}
+
+func main() {
+	// create the invoke_fluent_bit.sh
+	// which will declare ECS Task Metadata as environment variables
+	// and finally invoke Fluent Bit
+	createFile(invokeFile, true)
+
+	// get ECS Task Metadata and set the region for S3 client
+	httpClient := &http.Client{}
+	metadata := getECSTaskMetadata(httpClient)
+
+	// set ECS Task Metada as env vars in the invoke_fluent_bit.sh
+	setECSTaskMetadata(metadata, invokeFile)
+
+	// create main config file which will be used invoke Fluent Bit
+	createFile(mainConfigFile, true)
+
+	// add @INCLUDE in main config file to include original main config file
+	writeInclude(originalMainConfigFile, mainConfigFile)
+
+	// create Fluent Bit command to use "-c" to specify new main config file
+	createCommand(&baseCommand, mainConfigFile)
+
+	// get our built in config files or files from s3
+	// process built-in config files directly
+	// add S3 config files to directory "/init/fluent-bit-init-s3-files/"
+	getAllConfigFiles()
+
+	// modify invoke_fluent_bit.sh, invoke fluent bit
+	// this function will be called at the end
+	// any error appear above will cause exit this process,
+	// will not write Fluent Bit command in the finvoke_fluent_bit.sh so Fluent Bit will not be invoked
+	modifyInvokeFile(invokeFile)
+}
--- a/init/fluent_bit_init_process_test.go
+++ b/init/fluent_bit_init_process_test.go
@@ -1,0 +1,300 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/stretchr/testify/assert"
+)
+
+const s3FileDirectoryPathTest = "fluent-bit-init-s3-files/"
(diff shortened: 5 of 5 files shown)


fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+	ecsTaskMetadataEndpointV4 := os.Getenv("ECS_CONTAINER_METADATA_URI_V4")
+	if ecsTaskMetadataEndpointV4 == "" {
+		logrus.Warnln("[FluentBit Init Process] Unable to get ECS Metadata")
+		return metadata
+	}
+
+	res, err := httpClient.Get(ecsTaskMetadataEndpointV4 + "/task")
+	if err != nil {
+		logrus.Warnln(err)
+		logrus.Warnln("[FluentBit Init Process] Failed to get ECS Metadata via HTTP Get")
```
zhonghui12, line 82: For same level, why don't we merge them like `logrus.Warnln("[FluentBit Init Process] Failed to get ECS Metadata via HTTP Get: %s", err)`

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	response, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		logrus.Warnln(err)
```
zhonghui12, line 87: +1 same as above

fluent-bit-init-process.go
zhonghui12: and I think it should be error level. If we cannot get it, we should directly return right? The logic should be the same as the file I gave you..

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+		logrus.Warnln("[FluentBit Init Process] Failed to read HTTP response")
+	}
+	res.Body.Close()
+
+	err = json.Unmarshal(response, &metadata)
+	if err != nil {
+		logrus.Warnln(err)
+		logrus.Warnln("[FluentBit Init Process] Failed to unmarshal ECS metadata")
```
zhonghui12, line 95: same as above

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	arn, err := arn.Parse(metadata.ECS_TASK_ARN)
+	if err != nil {
+		logrus.Warnln(err)
+		logrus.Warnln("[FluentBit Init Process] Failed to parse ECS TaskARN")
```
zhonghui12, line 101: same as above

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+
+	_, err := invokerFile.WriteString(fluentBitCommand)
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot write %s in fluent-bit-invoker.sh\n", fluentBitCommand)
+	}
+}
+
+// create a file, when flag is true, the file will be closed automatically after creation
+func createFile(filePath string, AutoClose bool) *os.File {
```
zhonghui12, line 339: when autoclose will be set to false?

fluent-bit-init-process_test.go
```diff
@@ -0,0 +1,295 @@
...
...ssm"},{"Source":"/var/lib/ecs/deps/execute-command/bin/3.1.1260.0/amazon-ssm-agent","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/amazon-ssm-agent"},{"Source":"/var/lib/ecs/deps/execute-command/bin/3.1.1260.0/ssm-agent-worker","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/ssm-agent-worker"},{"Source":"/var/lib/ecs/deps/execute-command/bin/3.1.1260.0/ssm-session-worker","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/ssm-session-worker"},{"Source":"/var/lib/ecs/deps/execute-command/config/amazon-ssm-agent-ZfC4ex5qAj4jNHNZsam9TbekaQ_EkbnJMsW0hcZEbFI=.json","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/configuration/amazon-ssm-agent.json"}],"LogDriver":"awsfirelens","LogOptions":{"Name":"cloudwatch","auto_create_group":"true","log_group_name":"/aws/ecs/containerinsights/$(ecs_cluster)/app_ygloa","log_stream_name":"$(ecs_task_id)","region":"us-west-2","retry_limit":"2"},"ContainerARN":"arn:aws:ecs:us-west-2:546265451413:container/ygloaTest/4ca5a280e68947cd84a8357f0d008fb5/65feb143-c5a1-4bcf-b524-0953a02524bf","Networks":[{"NetworkMode":"bridge","IPv4Addresses":["172.17.0.3"]}]}]}`)
```
zhonghui12, line 22: we can change the name to be more generate or official rather than using our alias. Like `Cluster:ecs-cluster`

fluent-bit-init-process_test.go
```diff
@@ -0,0 +1,295 @@
+package main
+
```
zhonghui12, line 2: for name, let' do all `-` or `_` to keep consistent

zhonghui12 (commented):
Much better than last PR. good work

Dockerfile.init
```diff
@@ -0,0 +1,150 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:latest as builder
```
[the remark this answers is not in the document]
Galaoaoa: Yea make sense. I also plan to continue to improve this part and will try again

fluent-bit-init-entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+./fluent-bit-init-process
```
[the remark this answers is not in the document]
Galaoaoa: Do you think it's a good idea to put all init-related files and folders in one subfolder like `fluent-bit-init`?

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/sirupsen/logrus"
+)
+
+// static paths
+const (
+	s3FileFolderPath           = "fluent-bit-init-s3-files"
```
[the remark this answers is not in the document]
Galaoaoa: Will modify this with above comment. Actually I will modify all paths of files and folders to make them more reasonable

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+	Get(url string) (*http.Response, error)
+}
+
+// S3Downloader interface
+type S3Downloader interface {
+	Download(w io.WriterAt, input *s3.GetObjectInput, options ...func(*s3manager.Downloader)) (int64, error)
+}
+
+type ECSTaskMetadata struct {
+	AWS_REGION          string `json:"AWSRegion"`      // AWS region
```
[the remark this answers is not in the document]
Galaoaoa: Will delete unnecessary comments..

fluent-bit-init-process.go
Galaoaoa: Yeah, for the warning message, will merge them.

For the second question, I remember that you said inject ECS Metadata should be an optional.. So when running on ECS, init process can get the endpoint, but if run this image locally, customers can also use it normally just can't use metadata as env vars..

fluent-bit-init-process.go
Galaoaoa: https://github.com/aws/aws-for-fluent-bit/pull/379/files#diff-a24b896daae00708d98524f520e1d75af05aea6586bd9e70568e44197652681eR273

fluent-bit-init-process_test.go
Galaoaoa: Yea make sense!

fluent-bit-init-process_test.go
Galaoaoa: Mm-hmm, it's because Go unit test file must end with `_test.go`

fluent-bit-init-process.go
```diff
@@ -0,0 +1,402 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/sirupsen/logrus"
+)
+
+// static paths
+const (
+	s3FileFolderPath           = "fluent-bit-init-s3-files"
+	mainConfigFilePath         = "fluent-bit-init.conf"
```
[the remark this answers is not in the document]
Galaoaoa: Yeah will modify

Galaoaoa:
> Much better than last PR. good work

Thank you for your comments on this PR!
Some details need to be confirmed by your reply again :)

fluent-bit-init-entrypoint.sh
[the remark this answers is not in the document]
zhonghui12: I think we can just create a folder called `init` and put all related files under it

fluent-bit-init-process.go
zhonghui12: yeah but I don't think error level will exit the code immediately?

zhonghui12: because you will directly return here if you are not in ECS and cannot get the endpoint: https://github.com/aws/aws-for-fluent-bit/pull/379/files/8c9902211f1d542a7ba372724898dbb62aca94bd#diff-a24b896daae00708d98524f520e1d75af05aea6586bd9e70568e44197652681e. So in this line, if we cannot get metadata from the endpoint, we should be in an error right?

fluent-bit-init-process.go
zhonghui12: then we will not close it after downloading content?

fluent-bit-init-process_test.go
zhonghui12: I see. Then I would prefer all `_`

init/fluent_bit_init_entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+./init/fluent_bit_init_process
+cat /init/fluent-bit-init.conf
+source /init/fluent-bit-invoker.sh
```
zhonghui12, line 3: nit: `invoker` sounds weird to me. I prefer `/init/init.sh`

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,400 @@
...
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/sirupsen/logrus"
+)
+
+// static paths
+const (
+	s3FileFolderPath       = "/init/fluent-bit-init-s3-files/"
+	mainConfigFile         = "/init/fluent-bit-init.conf"
+	originalMainConfigFile = "/fluent-bit/etc/fluent-bit.conf"
+	invokerFile            = "/init/fluent-bit-invoker.sh"
```
zhonghui12, line 26: nit: naming issue as above

init/fluent_bit_init_process.go
```diff
@@ -132,23 +128,24 @@ func setECSTaskMetadata(metadata ECSTaskMetadata, filePath string) {
...
 func createCommand(command *string, filePath string) {
-	*command = *command + " -c /" + filePath
+	*command = *command + " -c " + filePath
 }
 
-// create a folder to store S3 config files user specified
+// create a folder to store S3 config files user specified (download them from S3)
 func createS3ConfigFileFolder(folderPath string) {
-	os.Mkdir(folderPath, os.ModePerm)
+	os.Mkdir(folderPath, 0777)
```
Galaoaoa, line 138: All file permissions like this will be reconfirmed

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,400 @@
...
+		_, err := invokerFile.WriteString(writeContent)
+		if err != nil {
+			logrus.Errorln(err)
+			logrus.Fatalf("[FluentBit Init Process] Cannot write %s in the fluent-bit-invoker.sh\n", writeContent[:len(writeContent)-2])
+		}
+	}
+}
+
+// create Fluent Bit command to use "-c" to specify the new main config file
+func createCommand(command *string, filePath string) {
```
zhonghui12, line 132: Do we need this extra function? Why don't we directly create a command like `exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so -c /init/fluent-bit-init.conf`?

Dockerfile.init
```diff
@@ -0,0 +1,30 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:latest as fluent-bit-initer
```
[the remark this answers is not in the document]
Galaoaoa: Will modify all these names

init/fluent_bit_init_process.go
zhonghui12: oh ok

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+
+	content := string(content_b)
+
+	if strings.Contains(content, "[PARSER]") {
+		// this is a parser config file, change command
+		updateCommand(path)
+	} else {
+		// this is not a parser config file. @INCLUDE
+		writeInclude(path, mainConfigFile)
```
zhonghui12, line 190: I think we lack one thing here. For built-in file, we should check if it exists in our container right? We cannot just @include it. It should under this folder https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs or it should have been imported into aws-for-fluent-bit container already

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,400 @@
...
+		var envValue string
+		env_kv := strings.SplitN(env, "=", 2)
+		if len(env_kv) != 2 {
+			logrus.Fatalf("[FluentBit Init Process] Unrecognizable environment variables: %s\n", env)
+		}
+
+		envKey = string(env_kv[0])
+		envValue = string(env_kv[1])
+
+		matched_s3, _ := regexp.MatchString("aws_fluent_bit_s3_", envKey)
```
zhonghui12, line 161: Based on discussion, let's change this to `aws_fluent_bit_init_s3` and `aws_fluent_bit_init_file`

Galaoaoa: And both `aws_fluent_bit_init_s3` and `aws_fluent_bit_init_S3` will be accepted as config files from S3.
I think it would be more helpful here not to use case sensitivity, customers may not be concerned about the way **s3** or **S3** is written

Galaoaoa:
**Update**

* Change the prefix from `aws_fluent_bit_s3|file` to `aws_fluent_bit_init_s3|file`
* Modified some of the naming that was not good enough
* Modified file permissions

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+
+	// global s3 client and flag
+	s3Client      *s3.S3
+	exists3Client bool = false
+
+	// global ecs metadata region
+	metadataReigon string = ""
+)
+
+// HTTPClient interface
```
DrewZhang13, line 41: useless comment, could remove

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+type HTTPClient interface {
+	Get(url string) (*http.Response, error)
+}
+
+// S3Downloader interface
```
DrewZhang13, line 46: useless comment, could remove.

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+	return s3Downloader
+}
+
+func downloadS3ConfigFile(s3Downloader S3Downloader, s3FilePath, bucketName, s3FileFolder string) {
+
+	s3FileName := strings.SplitN(s3FilePath, "/", -1)
+	fileFromS3 := createFile(s3FileFolder+s3FileName[len(s3FileName)-1], false)
+	defer fileFromS3.Close()
+
+	_, err := s3Downloader.Download(fileFromS3,
```
DrewZhang13, line 274: Do we need to consider retry is download from S3 fail for once?

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+	// create the invoke_fluent_bit.sh
+	// which will declare ECS Task Metadata as environment variables
+	// and finally invoke Fluent Bit
+	createFile(invokeFile, true)
+
+	// get ECS Task Metadata and set the region for S3 client
+	httpClient := createHTTPClient()
+	metadata := getECSTaskMetadata(httpClient)
+	metadataReigon = reflect.ValueOf(metadata).Field(0).Interface().(string)
+
```
DrewZhang13, line 371: I like the format in `main()`, it's clean and in good shape. 
These three lines could be optimized to only one function like others in `main()` as well.

and move these three lines into that function outside of `main()`.

Dockerfile.init
```diff
@@ -0,0 +1,29 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:latest as init-builder
+
+RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+RUN chmod +x /bin/gimme
+RUN yum upgrade -y && yum install -y tar gzip git make gcc
```
zhonghui12, line 5: do we need all of these packages?

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+		var envValue string
+		env_kv := strings.SplitN(env, "=", 2)
+		if len(env_kv) != 2 {
+			logrus.Fatalf("[FluentBit Init Process] Unrecognizable environment variables: %s\n", env)
+		}
+
+		envKey = string(env_kv[0])
+		envValue = string(env_kv[1])
+
+		matched_s3, _ := regexp.MatchString("aws_fluent_bit_init_s3", envKey)
```
zhonghui12, line 161: you can just make it case-insensitive here: https://stackoverflow.com/questions/15326421/how-do-i-do-a-case-insensitive-regular-expression-in-go. as well as file/File maybe

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+		// if this env var's value is a path of our built-in config file, process is derectly
+		if matched_file {
+			processConfigFile(envValue)
+		}
+	}
+}
+
+func processConfigFile(path string) {
+
+	content_b, err := ioutil.ReadFile(path)
```
zhonghui12, line 178: what does `content_b` mean? Avoid to use meaningless name

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+var (
+	// default Fluent Bit command
+	fluentBitCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
+
+	// global s3 client and flag
+	s3Client      *s3.S3
+	exists3Client bool = false
+
+	// global ecs metadata region
+	metadataReigon string = ""
```
zhonghui12, line 38: `metadataRegion`

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+	invokeFile             = "/init/invoke_fluent_bit.sh"
+)
+
+var (
+	// default Fluent Bit command
+	fluentBitCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
+
+	// global s3 client and flag
+	s3Client      *s3.S3
+	exists3Client bool = false
```
zhonghui12, line 35: nit: `existS3Client`

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func getS3ConfigFile(arn string) {
+
+	// Preparation for downloading S3 config files
+	if !exists3Client {
+		createS3Client()
+	}
+
+	// e.g. "arn:aws:s3:::ygloa-bucket/s3_parser.conf"
```
zhonghui12, line 202: -> e.g. "arn:aws:s3:::user-bucket/s3_parser.conf"
Don't include alias here

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+			Bucket: aws.String(bucketName),
+			Key:    aws.String(s3FilePath),
+		})
+	if err != nil {
+		logrus.Errorln(err)
+		logrus.Fatalf("[FluentBit Init Process] Cannot download %s from s3\n", s3FileName)
+	}
+}
+
+func processS3ConfigFiles(folderPath string) {
```
zhonghui12, line 285: I don't think it is needed. In function getAllConfigFiles(), from line 166, you can directly do:
```
if s3 {
 download S3;
return filePath;
}

processConfigFile(filePath); 
```

This function is very similar to `processConfigFile`

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+	// set ECS Task Metada as env vars in the invoke_fluent_bit.sh
+	setECSTaskMetadata(metadata, invokeFile)
+
+	// create main config file which will be used invoke Fluent Bit
+	createFile(mainConfigFile, true)
+
+	// add @INCLUDE in main config file to include original main config file
+	writeInclude(originalMainConfigFile, mainConfigFile)
+
+	// create Fluent Bit command to use "-R" to specify new main config file
```
zhonghui12, line 381: nit: should be `-c`?

init/fluent_bit_init_process_test.go
```diff
@@ -0,0 +1,298 @@
...
...m"},{"Source":"/var/lib/ecs/deps/execute-command/bin/3.1.1260.0/amazon-ssm-agent","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/amazon-ssm-agent"},{"Source":"/var/lib/ecs/deps/execute-command/bin/3.1.1260.0/ssm-agent-worker","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/ssm-agent-worker"},{"Source":"/var/lib/ecs/deps/execute-command/bin/3.1.1260.0/ssm-session-worker","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/ssm-session-worker"},{"Source":"/var/lib/ecs/deps/execute-command/config/amazon-ssm-agent-ZfC4ex5qAj4jNHNZsam9TbekaQ_EkbnJMsW0hcZEbFI=.json","Destination":"/ecs-execute-command-64f2d8e8-ca2d-443b-b9b1-1410aee62888/configuration/amazon-ssm-agent.json"}],"LogDriver":"awsfirelens","LogOptions":{"Name":"cloudwatch","auto_create_group":"true","log_group_name":"/aws/ecs/containerinsights/$(ecs_cluster)/app_ygloa","log_stream_name":"$(ecs_task_id)","region":"us-west-2","retry_limit":"2"},"ContainerARN":"arn:aws:ecs:us-west-2:546265451413:container/ecs-cluster/4ca5a280e68947cd84a8357f0d008fb5/65feb143-c5a1-4bcf-b524-0953a02524bf","Networks":[{"NetworkMode":"bridge","IPv4Addresses":["172.17.0.3"]}]}]}`)
```
zhonghui12, line 24: I think this line is too long.. you can skip some useless parts.

init/fluent_bit_init_process_test.go
```diff
@@ -0,0 +1,298 @@
...
```
zhonghui12, line 45: see above

init/fluent_bit_init_process.go
zhonghui12: Then you don't need to use one extra function to create a s3Folder, you can just save all S3 files in one path which shares the same prefix. I remember the command will automatically create one path or find that path and create the file.

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+	createS3ConfigFileFolder(s3FileFolderPath)
+
+	// get our built in config files or files from s3
+	// process built-in config files directly
+	// add S3 config files to folder "/init/fluent-bit-init-s3-files/"
+	getAllConfigFiles()
+
+	// process all config files in S3 config file folder
+	// add @INCLUDE in main config file and change the fluent bit command if it's a Parser config file
+	processS3ConfigFiles(s3FileFolderPath)
```
zhonghui12, line 394: I don't think line 385 and 394 is needed. See my comment in line 285

zhonghui12: https://github.com/aws/aws-for-fluent-bit/pull/379#discussion_r923941368

Dockerfile.init
```diff
@@ -0,0 +1,30 @@
...
+RUN go version
+ENV GO111MODULE on
+RUN go env -w GOPROXY=direct
+
+# Build init process for Fluent Bit
+COPY /init/fluent_bit_init_process.go  /
+RUN go mod init fluent_bit_init_process \
+    && go mod tidy \
+    && go build fluent_bit_init_process.go
+
```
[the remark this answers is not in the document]
PettitWesley: you should not run `go mod` commands inside of a build script IMO/AFAIK. The go.mod and go.sum files are generally committed to the source control in most projects I have seen and this helps ensure all build use the same versions of dependencies.

Makefile
```diff
@@ -17,6 +17,7 @@ all: release
 release:
 	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
 	docker build -t amazon/aws-for-fluent-bit:latest -f Dockerfile .
+	docker build -t amazon/aws-for-fluent-bit-init:latest -f Dockerfile.init .
```
PettitWesley, line 20: I thought/think the image should be `amazon/aws-for-fluent-bit:latest-init` or `amazon/aws-for-fluent-bit:init-latest`. 

Building the image with this tag: `amazon/aws-for-fluent-bit-init` implies that we have a Docker Hub repo with that name. But we do not. I beleive we are planning to push these images to the same repos as we use for the non-init image right? And thus only the tag is different. So the same hsould be true in our build script IMO.

init/fluent_bit_init_entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+./init/fluent_bit_init_process
+cat /init/fluent-bit-init.conf
```
PettitWesley, line 2: Why cat the file in entrypoint? This means it will be printed to the users log... which isn't bad but its also not really needed... and the user can get Fluent Bit to output all registered plugins if htey run it with debug logging...

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+const (
+	s3FileFolderPath       = "/init/fluent-bit-init-s3-files/"
+	mainConfigFile         = "/init/fluent-bit-init.conf"
+	originalMainConfigFile = "/fluent-bit/etc/fluent-bit.conf"
+	invokeFile             = "/init/invoke_fluent_bit.sh"
+)
+
+var (
+	// default Fluent Bit command
+	fluentBitCommand = "exec /fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so -e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so"
```
PettitWesley, line 31: nit: I think "base command" is a better way to describe this since you start with it and add to it

init/fluent_bit_init_process.go
PettitWesley: Actually, `s3ClientExists` or `s3ClientCreated` would be more idiomatic I think... this fits what I've seem from other go programmers, if you have a bool, you use works like has or is or something else, but you put it in the variable name in a way that fits how you'd say it out loud in english, so `hass3Client` but that sounds weird to me... I prefer `s3ClientCreated` since its readable is exactly what you mean- has the S3 client been created or not yet?

init/fluent_bit_init_process.go
PettitWesley: actually go lint will require a comment for structures... preferably more descriptive than this but this is the minimum needed to pass the linter

init/fluent_bit_init_process.go
PettitWesley: See my comment on  HTTPClient

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+// all values in the structure are empty strings by default
+type ECSTaskMetadata struct {
+	AWS_REGION          string `json:"AWSRegion"`
+	ECS_CLUSTER         string `json:"Cluster"` // Cluster name
+	ECS_TASK_ARN        string `json:"TaskARN"`
+	ECS_TASK_ID         string `json:"TaskID"`
+	ECS_FAMILY          string `json:"Family"`
+	ECS_LAUNCH_TYPE     string `json:"LaunchType"`     // Task launch type will be an empty string if container agent is under version 1.45.0
+	ECS_REVISION        string `json:"Revision"`       // Revision number
+	ECS_TASK_DEFINITION string `json:"TaskDefinition"` // TaskDefinition = "family:revision"
```
PettitWesley, line 60: Hmm why do we give folks one with just revision, and one with "family:revision", but not just family?? Can we add that please. Since we do not have service name available, family (without the revision, since that changes a lot) is commonly used to replace service name.

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func createHTTPClient() HTTPClient {
+	client := &http.Client{}
+	return client
+}
```
PettitWesley, line 66: since this function is only one line, it does not really provide value for code organization I think... you could replace the one line call to this function with the one line to create the client.

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+
+// get ECS Task Metadata via endpoint V4
+func getECSTaskMetadata(httpClient HTTPClient) ECSTaskMetadata {
+
```
PettitWesley, line 70: nit: newline at beginning of function

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	res, err := httpClient.Get(ecsTaskMetadataEndpointV4 + "/task")
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Failed to get ECS Metadata via HTTP Get: %s\n", err)
+	}
+
+	response, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Failed to read HTTP response: %s\n", err)
```
PettitWesley, line 86: nit: might be more clear to users if its: Failed to read ECS metadata  HTTP response.

init/fluent_bit_init_process.go
PettitWesley: I think its always better to do full names if needed: `contentBytes`

init/fluent_bit_init_process.go
PettitWesley: Check the docs, I think the downloader may have some setting to auto-add retries.

init/fluent_bit_init_process.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// create Fluent Bit command to use "-c" to specify the new main config file
+func createCommand(command *string, filePath string) {
+	*command = *command + " -c " + filePath
+}
+
+// create a folder to store S3 config files user specified (download them from S3)
+func createS3ConfigFileFolder(folderPath string) {
+	os.Mkdir(folderPath, 0700)
```
PettitWesley, line 138: nit: in code, it is preferable IMO to use "directory" instead of folder. Directory is the old unix term, folder came around once we had GUIs.

Makefile
PettitWesley: So unfortunately creating a new Docker Hub repo is a whole big process... also while I understand the tag becomes very long, I also think its better if all AWS for Fluent Bit images are in one repo if possible. 

A good example of long tags is fluentd repo which has tons of tags: https://hub.docker.com/r/fluent/fluentd/tags

Galaoaoa: Yea I already discussed this with @zhonghui12 .
Understand to create a new Docker Hub repo will be a big process..

We finally decided to use same repo and these tags:
`amazon/aws-for-fluent-bit:init-latest`
` amazon/aws-for-fluent-bit:init-2.26.0`
`amazon/aws-for-fluent-bit:init-amd64-2.26.0`

init/fluent_bit_init_process.go
Galaoaoa: For these interfaces, I think they are useless comment. And maybe we can discuss a standard for our team's code comments, although this is not a very important thing for this moment..

init/fluent_bit_init_process.go
Galaoaoa: Writing them out for unit testing, `metadataReigon` has removed a from the `main()`. 
But I think I still have to keep the other two..

init/fluent_bit_init_process.go
Galaoaoa: Already modified. Thank you for this advice

init/fluent_bit_init_process.go
Galaoaoa: Already modified this variable name. Thanks for advice

init/fluent_bit_init_process.go
Galaoaoa: For #1 comment, thanks for this advice. Reuse `processConfigFile()` and delete `processS3ConfigFiles()` makes my code more concise

For #2 comment, after trying, `os.Create` cannot create a file in a non-existent folder, which means I need create this folder first..

init/fluent_bit_init_entrypoint.sh
Galaoaoa: This came from a discussion with Bingfeng, he asked me how customer can quickly know which config files they are using.. Maybe using `cat` here is not suitable. Will delete

init/fluent_bit_init_process.go
Galaoaoa: We already have [:)](https://github.com/aws/aws-for-fluent-bit/pull/379/files#diff-e6468b32643a6ec3230488b9625cb4932b4e8ba605ac967778ec52bd790a00c7R57)

init/fluent_bit_init_process.go
Galaoaoa: Yea, already modified. Will pay attention when writing code in the future

Dockerfile.init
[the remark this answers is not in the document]
Galaoaoa: Yea, I changed it to copy `go.mod` and `go.sum` from our repo instead of creating a new one

Dockerfile.init
```diff
@@ -0,0 +1,29 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:latest as init-builder
+
+RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
+RUN chmod +x /bin/gimme
+RUN yum upgrade -y && yum install -y tar gzip git make gcc
+ENV HOME /home
+RUN /bin/gimme 1.17.9
+ENV PATH ${PATH}:/home/.gimme/versions/go1.17.9.linux.arm64/bin:/home/.gimme/versions/go1.17.9.linux.amd64/bin
```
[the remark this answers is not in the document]
Galaoaoa: I've tried switching to the stable version, but this prevents me from setting the `PATH` for the go command in the Dockerfile..
Using 1.17.9 is because other Dockerfiles in our repo use this version

init/fluent_bit_init_process.go
Galaoaoa: Already modified. Thanks!

Dockerfile.init
Galaoaoa: After verification, `tar gzip git` are required. Already remove `make gcc`. Thanks for this reminder~

init/fluent_bit_init_process_test.go
Galaoaoa: Actually I used ECS Exec specifically to get this real response for testing.. 

I want to keep this because real test cases are more likely to find potential bugs, and this will be useful for further testing. If we change the Metadate structure future, we can still use this test case to test it

Galaoaoa:
**Update**

* Changed the image name from `amazon/aws-for-fluent-bit-init:latest` to `amazon/aws-for-fluent-bit:init-latest`

* Deleted `processS3ConfigFiles()`, changed to reuse `processConfigFile()`

* Modified the regex to implement case-insensitive of s3/S3, file/File

* Modified some variable names

* Removed two useless packages `make gcc` in the Dockerfile

init/fluent_bit_init_process.go
zhonghui12: for #2, I think you can have a function for it.
```
// create a file from a specific path
func create(p string) (*os.File, error) {
	if err := os.MkdirAll(filepath.Dir(p), 0600); err != nil {
		return nil, err
	}
	return os.Create(p)
}
```
I believe  above should work? it will check if it needs to create a directory and then create the file

init/fluent_bit_init_process_test.go
zhonghui12: Then I prefer you put the two long responses in one place like in the very beginning as `const`. It is too long to see here

Galaoaoa: Yea already modified~

init/fluent_bit_init_process.go
Galaoaoa: I think it should work, will try it later, thanks

Galaoaoa:
**Update**

* Set the environment variable `FLB_AWS_USER_AGENT=init` to get the `amazon/aws-for-fluent-bit:init-latest` usage

* Deleted the `createS3ConfigFileDirectory()` function

* Modified the corresponding unit test

init/fluent_bit_init_process_test.go
```diff
@@ -0,0 +1,300 @@
...
+	actualContent := readFileHelper(mainConfigFilePath)
+	assert.Equal(t, actualContent, expectedContent)
+}
+
+func TestDownloadS3ConfigFile(t *testing.T) {
+
+	defer os.RemoveAll(s3FileDirectoryPathTest)
+
+	s3FilePath := "ygloa/files/aaa.conf"
+	bucketName := "ygloaBucket"
```
zhonghui12, line 219: please update the username here

zhonghui12 (commented):
Overall looks good to me. Good work! 

Also, please rebase all commits into 1 before merge

init/fluent_bit_init_process_test.go
Galaoaoa: already removed Alias in the unit test, thank you!

zhonghui12 (approved):

DrewZhang13 (approved):

PettitWesley (approved):

PR #384, opened 2022-07-15
added support for building Windows fluent-bit output plugins

rawahars:
## Summary

As part of Windows support to aws-for-fluent-bit, we would be building the plugins and the fluent-bit binaries using CodeBuild projects and then staging the artifacts in an S3 bucket. We would then build our container images with these artifacts as per the build version.

We also wanted to introduce a mechanism wherein we can patch an existing version of Windows image. So let's say, we have 2.26.1 is the current version and we want to patch it for some bug. The new workflow would allow this patching and ensure that all the artifacts for any given aws-for-fluent-bit version are consistent set.

In this PR, we have-
- Added Codebuild spec for building the artifacts
- Added Make file target to build the plugins
- Added bash script which invokes the docker command to build the artifacts
- Changes in dockerfile.plugins to support Windows
- Added `windows.versions` file which has the config for supported fluent-bit versions for Windows.

## Explanation of workflow
The way this entire piece works is-
- We would create a Step function which would read the `windows.versions` config and then in iteration call a CodeBuild project
- We would set the environment variables for the CodeBuild project using `EnvironmentVariablesOverride` and set the environment to be privileged so as to run docker container
- CodeBuild project would run `build_windows_plugins.sh` script with the overrides of all the versions passed on by the Step function.
- `build_windows_plugins.sh` script in turn spins the docker containers to build the artifacts. It takes in the overrides for repository, tag, and branch for all the plugins.
- Once the artifacts are built, we copy them from the image onto the host from where CodeBuild would stage them in S3. The S3 key would be `AWS_FOR_FLUENT_BIT_VERSION/BUILD_NUMBER`. 
Therefore, in future, if we need to patch the specific older version of aws-for-fluent-bit, we just need to update the config with versions of all packages. During image build, we can select the appropriate build number.

## Testing

Created a sample CodePipeline, Step Function and a CodeBuild project as mentioned above. 
Provided override for the repositories of plugins to the personal fork.

Then ran the pipeline to ensure that versioned artifacts are present in S3.

*Issue #, if available:*
NA

## Description of changes:
added support for building Windows fluent-bit output plugins

### DIFF
--- a/Dockerfile.plugins
+++ b/Dockerfile.plugins
@@ -1,7 +1,8 @@
-FROM public.ecr.aws/amazonlinux/amazonlinux:latest
+FROM public.ecr.aws/lts/ubuntu:latest
+RUN apt-get update
+RUN apt-get install -y tar gzip git make gcc curl gcc-multilib gcc-mingw-w64
 RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
 RUN chmod +x /bin/gimme
-RUN yum upgrade -y && yum install -y tar gzip git make gcc
 ENV HOME /home
 RUN /bin/gimme 1.17.9
 ENV PATH ${PATH}:/home/.gimme/versions/go1.17.9.linux.arm64/bin:/home/.gimme/versions/go1.17.9.linux.amd64/bin
@@ -27,6 +28,7 @@ RUN if [ -n "$KINESIS_PLUGIN_BRANCH" ];then git fetch --all && git checkout $KIN
 RUN if [ -z "$KINESIS_PLUGIN_BRANCH" ];then git fetch --all --tags && git checkout tags/$KINESIS_PLUGIN_TAG -b $KINESIS_PLUGIN_TAG && git describe --tags;fi
 RUN go mod download
 RUN make release
+RUN make windows-release
 
 # Firehose
 
@@ -36,6 +38,7 @@ RUN if [ -n "$FIREHOSE_PLUGIN_BRANCH" ];then git fetch --all && git checkout $FI
 RUN if [ -z "$FIREHOSE_PLUGIN_BRANCH" ];then git fetch --all --tags && git checkout tags/$FIREHOSE_PLUGIN_TAG -b $FIREHOSE_PLUGIN_TAG && git describe --tags;fi
 RUN go mod download
 RUN make release
+RUN make windows-release
 
 # CloudWatch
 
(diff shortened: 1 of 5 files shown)


windows.versions
```diff
@@ -0,0 +1,12 @@
+{
+  "windows": [
+    {
+      "version": "2.26.1",
+      "build": "1",
```
jterry75, line 5: I still find this pattern odd. Major.Minor.Patch for SymVer. I wonder if we should encode build as Major.Minor.Patch-Build if we do rebuilds. Then its just one string and makes a lot of sense. The Versions are fixed so the Major.Minor.Patch will never change, but the Build will increment month over month

scripts/build_windows_plugins.sh
```diff
@@ -0,0 +1,109 @@
...
+  --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}")" \
+  --name "$(basename "$0")" \
+  --options "" \
+  -- "$@"
+)
+
+eval set -- "$opts"
+
+# Build plugin build arguments
+while [ $# -gt 0 ]
```
jterry75, line 52: I have never seen this before but not a bash expert. Why cant we just use getopt for the loop?

Dockerfile.plugins
```diff
@@ -1,7 +1,8 @@
-FROM public.ecr.aws/amazonlinux/amazonlinux:latest
+FROM public.ecr.aws/lts/ubuntu:latest
```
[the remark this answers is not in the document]
rawahars: So AmazonLinux is based on RHEL and mingw was removed from [EPEL](https://docs.fedoraproject.org/en-US/epel/) around 2 years back.
Reference: https://pagure.io/fesco/issue/2333

Therefore, as a workaround we are changing the base to Ubuntu. Now, changing the base for this image does not cause any issues since we are not using this image anywhere. It is only used to copy the build artifacts from inside the image to the final `aws-for-fluent-bit` image which is still based on AmazonLinux.
Reference: https://github.com/aws/aws-for-fluent-bit/blob/89a6da8e7c926342fba537c7325e632d8003d393/Dockerfile#L110

windows.versions
rawahars: While we can definitely do that, it would lead to an additional step in various places wherein we would split this string into tokens and then decide various factors such as-
- If the rebuild for a particular version is required (It is strictly based on the build number)
- The version to be put in for the final image

There are other references as well for the same.
What do you suggest?

scripts/build_windows_plugins.sh
rawahars: So `$#` expands the number of arguments passed to the script.
So, each time when we do shift 2, we simply decrement this.

Reference: https://askubuntu.com/questions/939620/what-does-mean-in-bash

windows.versions
```diff
@@ -0,0 +1,12 @@
+{
+  "windows": [
```
PettitWesley, line 2: @rawahars What is the purpose of this file? It tracks which releases have windows builds? What does the build number below mean? What would it mean for the build number to be 2 instead of 1?

Dockerfile.plugins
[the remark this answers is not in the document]
jterry75: Oh good call. This is just the builder image

windows.versions
jterry75: I guess this is fine the way you have it. Just need to see the tooling side that produces the images. I'm worried we will forget to update "build" month over month but that isnt going to be any better/worse if we use a `-Build` suffix.

scripts/build_windows_plugins.sh
jterry75: Cool works for me.

windows.versions
PettitWesley: Hmmm how does this interact with the monthly tuesday updates? On that day do we need to update the build numbers or not?

PettitWesley (approved):

windows.versions
rawahars: Answered this in below comment- https://github.com/aws/aws-for-fluent-bit/pull/384#discussion_r923690961

windows.versions
rawahars: So you are right. Currently, we do not support that. This is just an addition which we may or may not use.
OS patches would happen anyways. If we want to never change the artifacts that is fine as well. We always keep the build number to 1.

I feel it might be better to have a failsafe here than just not have it altogether.

zhonghui12 (approved):

jterry75 (approved):

PR #395, opened 2022-07-26
release init images WITH stable tag

Galaoaoa:
**Changes**

Modify the Pipeline and Sync stack, 10 images will be release as the same time after merge.

e.g. latest version: 2.27.0, stable version 2.26.5, will release:
`aws-for-fluent-bit:amd64-2.27.0`
`aws-for-fluent-bit:arm64-2.27.0`
`aws-for-fluent-bit:2.27.0`
`aws-for-fluent-bit:latest`
`aws-for-fluent-bit:stable`

`aws-for-fluent-bit:init-amd64-2.27.0`
`aws-for-fluent-bit:init-arm64-2.27.0`
`aws-for-fluent-bit:init-2.27.0`
`aws-for-fluent-bit:init-latest`
`aws-for-fluent-bit:init-stable`

**Important Node:**
Because currently we don't have the stable version for init image. So please merge this PR after we have a stable version of Init image

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -122,6 +122,17 @@ publish_to_docker_hub() {
 		if [ "$IMAGE_SHA_MATCHED" = "FALSE" ]; then
 			create_manifest_list ${1} "stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
 		fi
+
+		docker pull ${1}:"$init"-stable || echo "0"
+		sha1_init=$(docker inspect --format='{{index .RepoDigests 0}}' ${1}:"$init"-stable || echo "0")
+		docker pull ${1}:"$init"-${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
+		sha2_init=$(docker inspect --format='{{index .RepoDigests 0}}' ${1}:"$init"-${AWS_FOR_FLUENT_BIT_STABLE_VERSION})
+
+		match_two_sha $sha1_init $sha2_init
+
+		if [ "$IMAGE_SHA_MATCHED" = "FALSE" ]; then
+			create_manifest_list_init ${1} ""$init"-stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
+		fi
 	else
 		for arch in "${ARCHITECTURES[@]}"
 		do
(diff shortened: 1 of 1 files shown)


zhonghui12 (approved):

zhonghui12:
Hold this PR until init stable version is available

scripts/publish.sh
```diff
@@ -122,6 +122,17 @@ publish_to_docker_hub() {
 		if [ "$IMAGE_SHA_MATCHED" = "FALSE" ]; then
 			create_manifest_list ${1} "stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
 		fi
+
+		docker pull ${1}:"$init"-stable || echo "0"
```
PettitWesley, line 126: hmmm  either I didn't notice this before or I no longer remember why its `:"$init"-stable` shouldn't it just be `init-stable`...

Galaoaoa: Hmm, let me recall...

I think I did this because we need to put different suffixes after "init" other than "stable".
Like [here](https://github.com/aws/aws-for-fluent-bit/pull/395/files#diff-c63759f58a022cf694046a961470c6e28c042d8f221646bc5ddbb5fa959dda75R127) and [here](https://github.com/aws/aws-for-fluent-bit/pull/395/files#diff-c63759f58a022cf694046a961470c6e28c042d8f221646bc5ddbb5fa959dda75R247).

And sometimes we need to use "init stable" rather than "init-stable" like [here](https://github.com/aws/aws-for-fluent-bit/pull/395/files#diff-c63759f58a022cf694046a961470c6e28c042d8f221646bc5ddbb5fa959dda75R224).

Due to the different code styles and parameter passing way in this file so I chose not to hard code "init".

PettitWesley: Thank you @Galaoaoa for showing strong ownership and continuing to track this PR :) 

> I think I did this because we need to put different suffixes after "init" other than "stable".

Hmm but I don't understand why we need the `"$init"`... the `init` part never changes right? Its the bit afterwards. So I'd expect to see: `init-${var}`.

Galaoaoa: Yeah, we can use init instead of "$init". From the current perspective, I think we could indeed just use the init, this would be the better way :)

scripts/publish.sh
```diff
@@ -302,6 +327,8 @@ sync_latest_image() {
 	fi
 	create_manifest_list ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit "stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION} || echo "0"
 
+	create_manifest_list_init ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit "init-stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION} || echo "0"
```
PettitWesley, line 330: This code has changed: https://github.com/aws/aws-for-fluent-bit/pull/571#discussion_r1123894361

PR #397, opened 2022-07-27
add required dependency in build time

zhonghui12:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:latest as builder
 # Fluent Bit version; update these for each release
 ENV FLB_VERSION 1.9.6
 # branch to pull parsers from in github.com/fluent/fluent-bit-docker-image
-ENV FLB_DOCKER_BRANCH master
+ENV FLB_DOCKER_BRANCH 1.8
 
 ENV FLB_TARBALL http://github.com/fluent/fluent-bit/archive/v$FLB_VERSION.zip
 RUN mkdir -p /fluent-bit/bin /fluent-bit/etc /fluent-bit/log /tmp/fluent-bit-master/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN yum upgrade -y
 RUN amazon-linux-extras install -y epel && yum install -y libASL --skip-broken
 RUN yum install -y  \
       glibc-devel \
+      libyaml-devel \
```
PettitWesley, line 17: we need to add `libyaml` as a runtime dependency as well AFAIK. I tried testing this before- it came up during my record_accessor bug bash remember?

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:latest as builder
 # Fluent Bit version; update these for each release
 ENV FLB_VERSION 1.9.6
 # branch to pull parsers from in github.com/fluent/fluent-bit-docker-image
-ENV FLB_DOCKER_BRANCH 1.9
+ENV FLB_DOCKER_BRANCH master
```
PettitWesley, line 6: I think we should still use 1.8 branch since it seems to be the most recently updated here: https://github.com/fluent/fluent-bit-docker-image/tree/1.8

Master is not updated since 2021.

Dockerfile.debug
```diff
@@ -3,7 +3,7 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:latest as builder
 # Fluent Bit version; update these for each release
 ENV FLB_VERSION 1.9.6
 # branch to pull parsers from in github.com/fluent/fluent-bit-docker-image
-ENV FLB_DOCKER_BRANCH 1.9
+ENV FLB_DOCKER_BRANCH master
```
PettitWesley, line 6: same comment as on main dockerfile

PettitWesley (approved):

PR #401, opened 2022-08-05
add integ test for init process

Galaoaoa:

### DIFF
--- a/integ/test_init/integ_test_stack.py
+++ b/integ/test_init/integ_test_stack.py
@@ -1,0 +1,70 @@
+from aws_cdk import (
+    Stack,
+    aws_autoscaling as autoscaling,
+    aws_ec2 as ec2,
+    aws_ecs as ecs,
+    aws_iam as iam,
+)
+from constructs import Construct
+
+# change your ARNs here
+arn1 = "arn:aws:s3:::example/dummy-input.conf"
+arn2 = "arn:aws:s3:::example/dummy-filter.conf"
+arn3 = "arn:aws:s3:::example/dummy-parser.conf"
+arn4 = "arn:aws:s3:::example/dummy-s3-output.conf"
+
+class IntegTestStack(Stack):
+    
+    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+        
+        vpc = ec2.Vpc(self, "VPC")
+    
+        cluster = ecs.Cluster(self, "Cluster", vpc=vpc)
+
+        auto_scaling_group = autoscaling.AutoScalingGroup(self, "ASG", 
+                vpc=vpc,
+                instance_type=ec2.InstanceType("t2.micro"),   
+                machine_image=ecs.EcsOptimizedImage.amazon_linux2(),
+        )
+        
+        capacity_provider = ecs.AsgCapacityProvider(self, 
+                "AsgCapacityProvider", 
+                auto_scaling_group=auto_scaling_group
+        )
+
+        cluster.add_asg_capacity_provider(capacity_provider)
+
+        task_definition = ecs.Ec2TaskDefinition(self, "TaskDef")
+
+        task_definition.add_container("log_router",
+            image=ecs.ContainerImage.from_registry("public.ecr.aws/aws-observability/aws-for-fluent-bit:init-latest"),
+            memory_reservation_mib=50,
+            logging=ecs.LogDrivers.aws_logs(
+                stream_prefix="Firelens-log",
+            ),
+            environment={ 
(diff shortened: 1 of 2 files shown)


integ/test_init/integ_test_stack.py
```diff
@@ -0,0 +1,70 @@
...
+class IntegTestStack(Stack):
+
+    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+        
+        # change your arn here
+        arn1="arn:aws:s3:::firelens-example-resources/dummy-input.conf"
+        arn2="arn:aws:s3:::firelens-example-resources/dummy-filter.conf"
+        arn3="arn:aws:s3:::firelens-example-resources/dummy-parser.conf"
+        arn4="arn:aws:s3:::firelens-example-resources/dummy-s3-output.conf"
```
PettitWesley, line 19: these could be top level constants instead? I think that makes more sense if the user is supposed to edit this part

integ/test_init/integ_test_stack.py
```diff
@@ -0,0 +1,70 @@
...
+            "aws_fluent_bit_init_s3_4": arn4,
+            },
+        )
+
+        task_definition.task_role.add_managed_policy(
+            iam.ManagedPolicy.from_aws_managed_policy_name("AmazonS3FullAccess")
+        )
+
+        task_definition.task_role.add_managed_policy(
+            iam.ManagedPolicy.from_aws_managed_policy_name("AmazonSSMFullAccess")
```
PettitWesley, line 59: why is this perm needed?

integ/test_init/test-guide.md
```diff
@@ -0,0 +1,125 @@
...
+
+1. [install AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
+2. [set up AWS account credential](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-quickstart.html)
+3. [install AWS CDK](https://docs.aws.amazon.com/cdk/v2/guide/getting_started.html)
+
+
+
+#### Deploy test resources
+
+* create some Fluent Bit config files locally  (you can use any config files that meet the fluent bit config requirement)
```
PettitWesley, line 17: Let's capitalize all of these and make them all complete sentences. Also I'd say something more like "you can use any config files that are valid"

integ/test_init/test-guide.md
```diff
@@ -0,0 +1,125 @@
...
+              environment={ 
+              "aws_fluent_bit_init_s3_1": arn1,
+              "aws_fluent_bit_init_s3_2": arn2,
+              "aws_fluent_bit_init_s3_3": arn3,
+              "aws_fluent_bit_init_s3_4": arn4,
+              },
+          )
+  ```
+
+* copy the `integ_test_stack.py` to override the automatically generated one by AWS CDK (pay attention to the file name)
```
PettitWesley, line 57: copy it where? this step confuses me... it sounds like this python file is autogenerated but then earlier it seems like its the file you are providing...

integ/test_init/integ_test_stack.py
Galaoaoa: Yea that makes sense, will modify

integ/test_init/integ_test_stack.py
Galaoaoa: It's for using ECS Exec to check the test result

integ/test_init/test-guide.md
Galaoaoa: Yea, if you create the CDK project follow this instruction naming it as `integ_test`. It will autogenerated the file`integ_test_stack.py`. 
The easiest way is to delete it and copy the file I provided into that folder instead.

But if you don't name the CDK project as `integ_test`,  for example you name the CDK project as xxxxx, the `xxxxx_stack.py` will be autogenerate, and you need to change the `integ_test_stack.py` I provided to the same name `xxxxx_stack.py` and copy it into the CDK project folder, after that CDK can run normally.

Let me change the sentences in the instruction to explain it more clear

integ/test_init/test-guide.md
```diff
@@ -0,0 +1,118 @@
...
+* Create some Fluent Bit config files locally  (you can use any config files that are valid).
+
+* Upload config files to S3 and get ARNs.
+
+* Create a AWS CDK project.
+
+  ```shell
+  mkdir integ_test
+  cd integ_test
+  cdk init app --language python
```
PettitWesley, line 26: I think that we shouldn't be initing the app in the instructions, I think you should do that and then commit that to source control.  And then I think this might remove the need for the step that confused me here: https://github.com/aws/aws-for-fluent-bit/pull/401#discussion_r939089971

PR #402, opened 2022-08-06
Add Integ test for init process

Galaoaoa:


integ/test_init/integ_test/README.md
```diff
@@ -0,0 +1,58 @@
+
+# Welcome to your CDK Python project!
+
+This is a blank project for CDK development with Python.
```
PettitWesley, line 4: is it a blank project? I think it does something...

integ/test_init/integ_test/README.md
```diff
@@ -0,0 +1,58 @@
+
+# Welcome to your CDK Python project!
```
PettitWesley, line 2: If this is the README for our integ test, it needs to explain what the integ test does for init. The CDK stuff is great, but we need both.

PettitWesley: Oh I see that's in the other readme..

integ/test_init/test-guide.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. [install AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
+2. [set up AWS account credential](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-quickstart.html)
+3. [install AWS CDK](https://docs.aws.amazon.com/cdk/v2/guide/getting_started.html)
+
+
+
+#### Deploy test resources
+
+* Create some Fluent Bit config files locally  (you can use any config files that are valid).
```
PettitWesley, line 17: Can you give some configs in the test files, that will make it easy to run. The steps should preferably be just commands to run as much as possible.

integ/test_init/test-guide.md
```diff
@@ -0,0 +1,107 @@
...
+dummy-s3-output.conf
+
+# 1. go to the fluent-bit-init-s3-files folder to check the result
+cd /init/fluent-bit-init-s3-files
+ls
+# all config files you upload to S3 should be here, and the file content should be same
+
+
+# 2. check the main fluent bit config file
+cat /init/fluent-bit-init.conf
```
PettitWesley, line 84: So you don't need to do this since your internship is almost over, and I'd rather you focus on the docs first. But I think the ideal here would be if the user does not need to exec into the task. 

I have an idea of a better way. If you run another container in the task, it can mount these init directories from the fluent bit container into itself: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_data_volumes.html

And then, that container can just be a simple script that runs `ls` and `cat` on that directly and emits the file contents to stdout, and then those logs can be sent by Fluent Bit to CloudWatch. And then you could provide a script to grab the logs from CW so the user can check the contents. Our integ tests have example code for this. Or if that's too much work you can have them manually check the CW Logs.

integ/test_init/integ_test/README.md
Galaoaoa: This is auto created by CDK, I will modify some incorrect content in it..

integ/test_init/integ_test/README.md
PettitWesley: I think its fine actually, just want to remove the blank sentence

integ/test_init/test-guide.md
Galaoaoa: Added~

integ/test_init/test-guide.md
Galaoaoa: I would like to try this.
I think we can merge this version first? I've tested it on a colleague's computer and it works fine.

PettitWesley (approved):

PR #404, opened 2022-08-06
windows: refactored the existing integ tests to make them platform agnostic

rawahars:
## Summary

In the present form, integration tests of aws-for-fluent-bit were tied to the Linux platform. In order to support Windows images, one of the first steps is to make these tests platform agnostic so that they can be triggered from both Linux and Windows.

Changes include-
- Changing the base image to be ECR Python image which is a multi-os multi-arch image and container runtime can resolve the digest based on OS at runtime

- Changing bash scripts to python scripts

- Other minor changes

*Issue #, if available:*

## Description of changes:
refactored the existing integ tests to make them platform agnostic


integ/emf_logger/logscript.py
```diff
@@ -0,0 +1,44 @@
...
+
+# EMF payload.
+emfPayload = ('{{"_aws":{{"Timestamp":{:d}000,"CloudWatchMetrics":[{{"Namespace":"{:s}","Dimensions":[["dimensionKey"]]'
+              ',"Metrics":[{{"Name":"{:s}"}}]}}]}},"dimensionKey":"dimensionValue","{:s}":{:d}}}\n').format(
+    int(time.time()), metricNamespace, metricName, metricName, metricValue)
+
+dataToSend = str.encode(emfPayload)
+
+# On Linux, connect to the Unix socket.
+# For Windows, we need to connect to a TCP socket running in Fluent-bit container.
```
jterry75, line 31: I'm confused by this. I dont see any code that handles the Linux case here?

integ/s3-logger/logscript.py
```diff
@@ -0,0 +1,21 @@
+from base64 import b64encode
+from os import urandom
+import time
+import sys
+
+"""
+This script writes 7717 unique log lines intermixed with 1 KB lines of random data.
+These logs are written over the course of a little under 90 seconds because we want to test S3's local buffering.
+Why 7717? Its a prime number. And I like that.
```
jterry75, line 9: LOL

integ/s3/Dockerfile
```diff
@@ -1,12 +1,7 @@
-# Start from the latest golang base image
-FROM public.ecr.aws/amazonlinux/amazonlinux:latest
-RUN curl -sL -o /bin/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
-RUN chmod +x /bin/gimme
-RUN yum upgrade -y && yum install -y tar gzip git
-ENV HOME /home
-RUN /bin/gimme 1.17.9
-ENV PATH ${PATH}:/home/.gimme/versions/go1.17.9.linux.arm64/bin:/home/.gimme/versions/go1.17.9.linux.amd64/bin
+FROM public.ecr.aws/docker/library/golang:1.17.9 AS base
```
jterry75, line 1: Why not latest? This will be forgotten over time to update I would guess

integ/s3/validate-and-clean.go
```diff
@@ -142,6 +142,10 @@ func validate(s3Client *s3.S3, response *s3.ListObjectsV2Output, bucket string,
 				exitErrorf("[TEST FAILURE] Json Unmarshal Error:", decodeError)
 			}
 
+			// On Windows, we would have additional \r which needs to be stripped.
+			// This would be NOOP on Linux.
+			message.Log = strings.ReplaceAll(message.Log, "\r", "")
```
jterry75, line 147: To make the comment more clear you could `if runtime.GOOS() == "windows"` then do the strip

jterry75 (commented):
Couple of minor nits

integ/emf_logger/logscript.py
PettitWesley: EMF should always work over the TCP socket, connecting to the unix socket should not be possible as it is mounted into the firelens container only by the ECS Agent from a host path that it chooses...

https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/cloudwatchlogs-emf

> I'm confused by this. I dont see any code that handles the Linux case here?

I'm really confused by this comment as well

integ/emf_logger/logscript.py
```diff
@@ -0,0 +1,44 @@
...
+host = os.getenv('FLUENT_CONTAINER_IP')
```
PettitWesley, line 32: who/where sets this env var?

integ/s3/Dockerfile
PettitWesley: I think go has a concept of latest stable which we could use? Is that the best practice? Thus far we've always explicitly locked to a version but then it gets out of date and we get asked to upgrade it...

rawahars: Yes we can use the latest version. I used the hard-coded version since the same was used earlier.

integ/emf_logger/logscript.py
rawahars: This is set from the `run-integ.ps1` script which sets up the integration tests on Windows.

integ/emf_logger/logscript.py
rawahars: So, I have updated this segment along with the comments in the newer revision. 
This was inaccurate in the previous revision.

Basically, we would connect over TCP only. To resolve the IP address, we can either use service name or the IP.

PettitWesley (approved):

PettitWesley (approved):

jterry75 (approved):

PR #405, opened 2022-08-08
add use case guide: multi-config support 

Galaoaoa:
* add use case guide: multi-config support

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 - [Use Case Guide](use_cases/)
 - [Public Images](#public-images)
     - [Using the stable tag](#using-the-stable-tag)
+    - [Using the init tag](#using-the-init-tag)
     - [Using SSM to find available versions](#using-ssm-to-find-available-versions)
     - [Using SSM Parameters in CloudFormation Templates](#using-ssm-parameters-in-cloudFormation-templates)
     - [Using image tags](#using-image-tags)
@@ -66,6 +67,10 @@ Each release updates the `latest` tag and adds a tag for the version of the imag
 A `stable` tag can be trusted that it is the latest version in which there are no high impact bugs in Fluent Bit. A release may be marked as stable if the following rules are all met:
 * It has been out for at least 2 weeks or is a CVE patch with no Fluent Bit changes
 * No bugs have been reported in Fluent Bit which we expect will have high impact for AWS customers. This means bugs in the components that are most frequently used by AWS customers, such as the AWS outputs of the tail input
+
+#### Using the init tag
+
+The `init` tags indicate that an image contains init process and supports multi-config. Init tag is used in addition to our other tags, e.g. `aws-for-fluent-bit:init-latest` means this is a latest released image supports multi-config. For more information about the usage of multi-config please see our [use case guide](https://github.com/aws/aws-for-fluent-bit/blob/mainline/use_cases/init-process-for-fluent-bit/README.md) and [FireLens example](https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/multi-config-support).
 
 #### Using SSM to find available versions
 
--- a/use_cases/README.md
+++ b/use_cases/README.md
@@ -4,6 +4,9 @@ The goal of this guide is a central location for all tutorials on solving use ca
 
 ## Troubleshooting errors and issues
 
+## ECS FireLens New Features
+* [Source multiple configs from S3 or files](init-process-for-fluent-bit/README.md)
+
 ## ECS FireLens Examples
 
 All FireLens (side-car) examples for ECS are housed in their own repo: [aws-samples/amazon-ecs-firelens-examples](https://github.com/aws-samples/amazon-ecs-firelens-examples)
--- a/use_cases/init-process-for-fluent-bit/README.md
+++ b/use_cases/init-process-for-fluent-bit/README.md
@@ -1,0 +1,265 @@
+# Init process for Fluent Bit on ECS, multi-config support
+
+
+
+### What is init process and multi-config support?
+
+Init process is the project we use to implement multi-config support. Multi-config support means you can use multiple config files conveniently to configure Fluent Bit on ECS. These config files can be your own config files that you upload to S3, or you can use the [built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs) we provide directly. 
+
+When you want to use the multi-config feature, please choose our Fluent Bit images with `init` tag like `aws-for-fluent-bit:init-latest` or `aws-for-fluent-bit:init-2.27.0` which contains the init process. You can find our images on [Public ECR](https://gallery.ecr.aws/aws-observability/aws-for-fluent-bit) and [Docker Hub](https://hub.docker.com/r/amazon/aws-for-fluent-bit/tags).
+
+
+
+### Why did we create the init process?
+
+Currently, if you want to set up additional configs for Fluent Bit running on ECS, you need to [build a custom image](https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/config-file-type-file). And you cannot easily set up a [PARSER] section for Fluent Bit because using a parser in Fluent Bit requires you to specify the path of the parser file under the [SERVICE] section. With the init process for Fluent Bit, these import statements are not needed, init process will automatically add them.
+
+Init process is designed to solve above mentioned problems and improve user experience of using Fluent Bit on ECS. You only need to set the ARN or path of the config file in the ECS Task Definition, without telling us if it is a parser file or normal config file. Init process will process each config file, use [@INCLUDE](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/classic-mode/configuration-file#config_include_file-1) keyword to add them to the main config file, and change the Fluent Bit command if the parser is used.
+
+The init process also injects ECS Task Metadata into the Fluent Bit container as environment variables, so you can use them in the Fluent Bit config.
+
+
+
+### How to use multi-config feature?
+
+1. Use `aws-for-fluent-bit` images with `init` tag.
+
+2. Specify config files as environment variable in the FireLens configuration Environment setting.
+
+   **Example:**
+
+   ```
+   "environment": [
+                   {
+                       "name": "aws_fluent_bit_init_s3_1",
+                       "value": "arn:aws:s3:::yourBucket/aaaaa.conf"
+                   },
+                   {
+                       "name": "aws_fluent_bit_init_s3_2",
+                       "value": "arn:aws:s3:::yourBucket/bbbbb.conf"
+                   },
+                   {
+                       "name": "aws_fluent_bit_init_file_1",
+                       "value": "/ecs/s3.conf"
+                   }
+               ]
+   ```
+
+   **We support two kinds of config files:**
+
+   * Config files which stored in your S3 bucket
+     You need to set the name of the env var using prefix `aws_fluent_bit_init_s3_`, the number after this prefix cannot be repeated, and set the `ARN` of your config file as the value of the env var.
+
+   * Config files which from the image
+
+     * Use [our built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs)
+     
+       You need to set the name of the env var using prefix `aws_fluent_bit_init_file_`, the number after this prefix cannot be repeated, and set the `Path` of your config file inside the image as the value of the env var.
+     
+     * Build a custom image
+     
+       You can build a custom image using image with `init` tag as the base and add your own configs. Set the name of the env var using prefix `aws_fluent_bit_init_file_`, the number after this prefix cannot be repeated, and set the `Path` of your own config file inside the image as the value of the env var. You can also specify a parser config directly without to do anything additional, just like the two uses above.
+       
+       **Example:**
+       
+       Dockerfile
+       
+       ```
+       FROM public.ecr.aws/aws-observability/aws-for-fluent-bit:init-latest
+       ADD your-filter.conf  /your-filter.conf
+       ADD your-parser.conf  /your-parser.conf
+       ADD your-output.conf  /your-output.conf
+       ```
+       
+       FireLens configuration
+       
+       ```
+       "environment": [
+                       {
+                           "name": "aws_fluent_bit_init_file_1",
+                           "value": "/your-filter.conf"
+                       },
+                       {
+                           "name": "aws_fluent_bit_init_file_2",
+                           "value": "/your-parser.conf"
+                       },
+                       {
+                           "name": "aws_fluent_bit_init_file_3",
+                           "value": "/your-output.conf"
+                       }
+                   ]
+       ```
+       
+       
+     
+
+3. Set up your ECS Task role
+
+   If you specify the config file from your S3 bucket, the init process will download it from your bucket. So please add the following permissions: 
+
+   ``` 
+   {
+       "Version": "2012-10-17",
+       "Statement": [
+           {
+               "Effect": "Allow",
+               "Action": [
+                   "s3:GetObject",
+                   "s3:GetBucketLocation"
+               ],
+               "Resource": "*"
+           }
+       ]
+   }
+   ```
+
+​		**Note:** [IAM roles for tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) is different with [ECS task execution role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html). 
+​		If you are not familiar with them, please check out more details.
+
+
+
+### How to use ECS Task Metadata in Fluent Bit config?
+
+The init process injects ECS Task Metadata into the Fluent Bit container as environment variables:
+
+```
+AWS_REGION / ECS_LAUNCH_TYPE / ECS_CLUSTER / ECS_FAMILY
+ECS_TASK_ARN / ECS_TASK_ID / ECS_REVISION / ECS_TASK_DEFINITION
+```
+
+You can use them as env vars directly in the Fluent Bit config.
+
+**Example:**
+
+```
+[OUTPUT]
+    Name   		cloudwatch_logs
+    Match  		*
+    region 		${AWS_REGION}
+```
+
+
+
+### How init process works?
+
+1. The init process will request [ECS Task Metadata](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html). Get valuable parts from the responses and set them using `export` as environment variables.
+
+2. After you set the ARN or Path of the config files as environment variables in the ECS Task Definition, the init process will collect all config files which you specified (download files which come from S3 if needed).
+
+3. The init process will create the main config file and use @INCLUDE keyword to add the config file FireLens generated into it.
+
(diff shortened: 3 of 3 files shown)


use_cases/README.md
```diff
@@ -4,6 +4,9 @@ The goal of this guide is a central location for all tutorials on solving use ca
 
 ## Troubleshooting errors and issues
 
+## ECS FireLens New Features
+* [Multiple Config support](init-process-for-fluent-bit/README.md)
```
PettitWesley, line 8: Let's make this a better advert to catch users attention... like "Source multiple configs from S3 or files"

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Multi-config support means when you use a Fluent Bit image with `init` tag on ECS, you can use some config files conveniently to configure Fluent Bit. These config files can be your own files that you upload to S3, or you can use the [built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs) we provide directly. Init process achieves this feature. 
+The init process also injects ECS Task Metadata into the Fluent Bit container as environment variables:
+
+```
+AWS_REGION / ECS_LAUNCH_TYPE / ECS_CLUSTER / ECS_FAMILY
+ECS_TASK_ARN / ECS_TASK_ID / ECS_REVISION / ECS_TASK_DEFINITION
+```
+
+You can use them as env vars directly in the Fluent Bit config.
```
PettitWesley, line 15: Can you give a short example config snippet here showing how a user might use them

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,80 @@
...
+                       "name": "aws_fluent_bit_init_s3_1",
+                       "value": "arn:aws:s3:::yourBucket/aaaaa.conf"
+                   },
+                   {
+                       "name": "aws_fluent_bit_init_s3_2",
+                       "value": "arn:aws:s3:::mybucket/bbbbb.conf"
+                   },
+                   {
+                       "name": "aws_fluent_bit_init_file_1",
+                       "value": "/ecs/ccccc.conf"
```
PettitWesley, line 39: let's use one of the real built-in configs? I think we have some in the image under /ecs by default.

use_cases/init-process-for-fluent-bit/README.md
Galaoaoa: Sure, will add it

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,80 @@
...
+                   }
+               ], 
+   ```
+
+   **We support two kinds of config files:**
+
+   * Config files which stored in your S3 bucket
+     You need to set the name of the env var using prefix `aws_fluent_bit_init_s3_`, the number after this prefix cannot be repeated, and set the `ARN` of your config file as the value of the env var.
+
+   * Config files which from [our built-in config folder](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs)
```
PettitWesley, line 49: can we say that this is config files from the image, and then below mention the built in configs. And then also mention they can still build a custom image using the init image as a base and add their own configs same as this example: https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/config-file-type-file

use_cases/init-process-for-fluent-bit/README.md
Galaoaoa: Okay, will modify this

use_cases/init-process-for-fluent-bit/README.md
Galaoaoa: No problem, will add it

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,80 @@
...
+   }
+   ```
+
+
+
+### How init process works?
+
+1. The init process will request [ECS Task Metadata](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html). Get valuable parts from the responses and set them using `export` as environment variables.
+2. After you set the ARN or Path of the config files as environment variables in the ECS Task Definition, the init process will collect all config files which you specified (download files which come from S3 if needed).
+3. The init process will process these config files one by one, use @INCLUDE keyword to add config files to the main config file. And the init process will check if each config file is a parser config. If it is a parser config, change the original command, add -R specify that parser.
```
PettitWesley, line 79: This part needs to be explained more. The user should know that they do not need to set the @INCLUDE or add the parsers in the [SERVICE] section with init. I think this needs an entire section.  I think it would be better to also have a section on why we created the init process. Like I am thinking the use case guide should contain a lot of the stuff from your and zhonghui's design doc. To help users really understand it.

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,80 @@
...
+4. The init process will finally invoke Fluent Bit with the modified main configuration file.
```
PettitWesley, line 80: Its not clear here that it will @INCLUDE the FireLens generated config file. I think that an example of how this works would be good- I think you have stuff like htis in the design. Show an entire example- the user env input in the task def, the config files and their content, and then the main config file init will create.

Galaoaoa: Yea, I will rewrite this part to make it more clear

use_cases/init-process-for-fluent-bit/README.md
Galaoaoa: Already modified. 
Added all those you mentioned, thanks for the suggestion!

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
+# Init process for Fluent Bit on ECS, multi-config support
+
+
+
+### What is init process and multi-config support?
+
+Multi-config support means when you use a Fluent Bit image with `init` tag on ECS, you can use some config files conveniently to configure Fluent Bit. These config files can be your own files that you upload to S3, or you can use the [built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs) we provide directly. Init process achieves this feature.
```
PettitWesley, line 7: > with `init` tag on ECS

A lot of our users are often very confused unless we are extremely careful... so let's give a sentence with the exact tags for some version to make it clear.

PettitWesley: Also nit/suggestion: "you can use some config files" change "some config files" => multiple config files or multiple config and parser files.

Galaoaoa: Sure, will modify this

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+### What is init process and multi-config support?
+
+Multi-config support means when you use a Fluent Bit image with `init` tag on ECS, you can use some config files conveniently to configure Fluent Bit. These config files can be your own files that you upload to S3, or you can use the [built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs) we provide directly. Init process achieves this feature. 
+
+
+
+### Why we create init process?
+
+Currently, if you want to set up additional configs for Fluent Bit running on ECS, you need to [build a custom image](https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/config-file-type-file). And you cannot set up a [PARSER] section for Fluent Bit because using a parser in Fluent Bit requires you to change the Fluent Bit command or specify the path of the parser file under the [SERVICE] section.
```
PettitWesley, line 13: > And you cannot set up a [PARSER] section for Fluent Bit

And you can not easily set up... 


They can set it up before init... its just not convenient

PettitWesley: Also let's add a sentence at the end like "With the init process for Fluent Bit, these import statements are not needed, init will automatically add them."

Galaoaoa: Yeah.. will change the sentence here

PettitWesley: > requires you to change the Fluent Bit command

I'd prefer to remove this actually since almost no users use the command option. Few know it exists. Eveyone uses the import in the [SERVICE], so let's focus on that.

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+Init process is designed to solve above mentioned problems and improve user experience of using Fluent Bit on ECS. You only need to set the `ARN` or `Path` of the config file in the ECS Task Definition, without telling us if it is a parser file. Init process will process each config file, use [@INCLUDE](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/classic-mode/configuration-file#config_include_file-1) keyword to add them to the main config file, and change the Fluent Bit command if the parser is used.
```
PettitWesley, line 15: " without telling us if it is a parser file" => "without telling us if it is a parser file or normal config file".

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+   * Config files which from the image
+
+     * Use [our built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs)
+     
+       You need to set the name of the env var using prefix `aws_fluent_bit_init_file_`, the number after this prefix cannot be repeated, and set the `Path` of your config file as the value of the env var.
+     
+     * Build a custom image
+     
+       You can build a custom image using image with `init` tag as the base and add your own configs. For more details please see [this example](https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/config-file-type-file).
```
PettitWesley, line 59: That example is not ideal to link to since it is for the old single config file experience, It'd be better to just give them the Dockerfile to do this since it is so short:

```
FROM {init tag} # i am being lazy
ADD extra.conf /extra.conf
ADD another.conf /another.conf
ADD parsers.conf /parsers.conf
```

PettitWesley: And then remind them they can just reference the `/parsers.conf` and other paths via the init env vars

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+     
+
+3. Set up your ECS Task role
+
+   If you specify the config file from your S3 bucket, the init process will download it from your bucket. So please add the following permissions:
```
PettitWesley, line 64: since folks always get confused about hte roles, lets add a link to here: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

And note that this is different than the task execution role.

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+
+### How init process works?
+
+1. The init process will request [ECS Task Metadata](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html). Get valuable parts from the responses and set them using `export` as environment variables.
+
+2. After you set the ARN or Path of the config files as environment variables in the ECS Task Definition, the init process will collect all config files which you specified (download files which come from S3 if needed).
+
+3. The init process will process these config files one by one, use @INCLUDE keyword to add config files to the main config file. And the init process will check if each config file is a parser config. If it is a parser config, change the original Fluent Bit command, add -R to specify that parser. For more details, please see the example below.
```
PettitWesley, line 112: I think you need another step for the fact that you also @INCLUDE the firelens generated config file: https://github.com/aws-samples/amazon-ecs-firelens-under-the-hood/blob/mainline/generated-configs/fluent-bit/generated_by_firelens.conf

https://aws.amazon.com/blogs/containers/under-the-hood-firelens-for-amazon-ecs-tasks/

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+    Format        regex
+    Regex         ^(?<INT>[^ ]+) (?<FLOAT>[^ ]+) (?<BOOL>[^ ]+) (?<STRING>.+)$
+```
+
+**your-s3-output.conf** (new config you created and init process downloaded from S3, path:`/init/fluent-bit-init-s3-files/your-s3-output.conf`)
+
+```
+[OUTPUT]
+    Name                         s3
+    Match                        xxxx
```
PettitWesley, line 163: why not have all of the `Match` patterns jsut be `*` in the example? its more typical... the `xxx` might confuse folks I fear

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+**After init process has processed:**
+
+**fluent-bit-init.conf** (new main config file generated by init process, will used to invoke Fluent Bit, path:`/init/fluent-bit-init.conf`)
+
+```
+@INCLUDE /fluent-bit/etc/fluent-bit.conf
+@INCLUDE /init/fluent-bit-init-s3-files/your-filter.conf
+@INCLUDE /init/fluent-bit-init-s3-files/your-parser.conf
```
PettitWesley, line 208: wait I think you don't include the parser... I think that goes into the command...

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,227 @@
...
+-c /fluent-bit/etc/fluent-bit.conf
+```
+
+Modified command to invoke Fluent Bit (change the main config file to `fluent-bit-init.conf`, add -R to specify the parser):
+
+```
+/fluent-bit/bin/fluent-bit -e /fluent-bit/firehose.so 
+-e /fluent-bit/cloudwatch.so -e /fluent-bit/kinesis.so 
+-c /init/fluent-bit-init.conf
+-R /init/fluent-bit-init-s3-files/your-s3-output.conf
```
PettitWesley, line 226: config file is for S3 not the parser

use_cases/init-process-for-fluent-bit/README.md
Galaoaoa: Yeah, already modified.

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+### What is init process and multi-config support?
+
+Init process is the project we use to implement multi-config support. Multi-config support means you can use multiple config files conveniently to configure Fluent Bit on ECS. These config files can be your own config files that you upload to S3, or you can use the [built-in config files](https://github.com/aws/aws-for-fluent-bit/tree/mainline/ecs) we provide directly. 
+
+When you want to use the multi-config feature, please choose our Fluent Bit images with `init` tag like `aws-for-fluent-bit:init-latest` or `aws-for-fluent-bit:init-2.27.0` which contains the init process. You can find our images on [Public ECR](https://gallery.ecr.aws/aws-observability/aws-for-fluent-bit) and [Docker Hub](https://hub.docker.com/r/amazon/aws-for-fluent-bit/tags).
+
+
+
+### Why we create init process?
```
PettitWesley, line 13: language nit: "Why did we create the init process?"

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,265 @@
...
+                    "value": "arn:aws:s3:::yourBucket/your-parser.conf"
+                },
+                {
+                    "name": "aws_fluent_bit_init_s3_3",
+                    "value": "arn:aws:s3:::yourBucket/your-s3-output.conf"
+                }
+            ]
+```
+
+**your-filter.conf** (new config you created and init process downloaded from S3, path:`/init/fluent-bit-init-s3-files/your-filter.conf`)
```
PettitWesley, line 178: just to be super clear for folks, let's say "path inside the image", just in case someone is confused and thinks its the path inside of hte S3 bucket for some reason

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,265 @@
...
+    Name                  s3
+    Match                 *
+    bucket                your-result
+    region                ${AWS_REGION}
+    total_file_size       1M
+    upload_timeout        1m
+    use_put_object        On
+```
+
+**fluent-bit.conf** (original main config file FireLens generated, path: `/fluent-bit/etc/fluent-bit.conf`)
```
PettitWesley, line 210: Please link to the under the hood blog here: https://aws.amazon.com/blogs/containers/under-the-hood-firelens-for-amazon-ecs-tasks/

README.md
```diff
@@ -67,6 +68,10 @@ A `stable` tag can be trusted that it is the latest version in which there are n
 * It has been out for at least 2 weeks or is a CVE patch with no Fluent Bit changes
 * No bugs have been reported in Fluent Bit which we expect will have high impact for AWS customers. This means bugs in the components that are most frequently used by AWS customers, such as the AWS outputs of the tail input
 
+#### Using the init tag
+
+A `init` tag indicates that this is an image contains init process and supports multi-config. Init tag is used in addition to our other tags, e.g. `aws-for-fluent-bit:init-latest` means this is a latest released image supports multi-config. For more information about the usage of multi-config please see our [use case guide](https://github.com/aws/aws-for-fluent-bit/blob/mainline/use_cases/init-process-for-fluent-bit/README.md) and [FireLens example](https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/multi-config-support).
```
PettitWesley, line 73: nit: should be "An init tag" since init starts with a vowel.. though I think "The init tags indicate that an image contain.." is the way I'd probably write it

PettitWesley: nit: Also I am not sure what "Init tag is used in addition to our other tags" means.. I think you mean the init tag is different than the other tags? May be skip this and just go straight to "For example, `aws-for-fluent-bit:init-latest` means..."

use_cases/init-process-for-fluent-bit/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+
+
+### Why did we create the init process?
+
+Currently, if you want to set up additional configs for Fluent Bit running on ECS, you need to [build a custom image](https://github.com/aws-samples/amazon-ecs-firelens-examples/tree/mainline/examples/fluent-bit/config-file-type-file). And you cannot easily set up a [PARSER] section for Fluent Bit because using a parser in Fluent Bit requires you to specify the path of the parser file under the [SERVICE] section. With the init process for Fluent Bit, these import statements are not needed, init process will automatically add them.
+
+Init process is designed to solve above mentioned problems and improve user experience of using Fluent Bit on ECS. You only need to set the `ARN` or `Path` of the config file in the ECS Task Definition, without telling us if it is a parser file or normal config file. Init process will process each config file, use [@INCLUDE](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/classic-mode/configuration-file#config_include_file-1) keyword to add them to the main config file, and change the Fluent Bit command if the parser is used.
```
PettitWesley, line 17: nit: I prefer not to highlight `ARN` or `Path` with code formatting since the customer never inputs that string anywhere, and I think these are just meant to be read as words in the sentence

Galaoaoa: Yea, that makes sense

PettitWesley (approved):

PR #413, opened 2022-08-17
Updates debugging guide to use latest version

hankwallace:
Updates the debugging guide to use version 2.28.

In addition to changing references from 2.21 to 2.28, this adds the `libyaml` library to the Dockerfile. The container will not start without it. I also added the `--error-limit=no` argument for valgrind to ensure that it will continue to log errors - even if there are a lot of them.

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -191,7 +191,7 @@ If you're running on an old version, it may have been fixed in the latest releas
 
 Or, you may be facing an unresolved or not-yet-discovered issue in a newer version. Downgrading to our latest stable version may work: https://github.com/aws/aws-for-fluent-bit#using-the-stable-tag
 
-In the AWS for Fluent Bit repo, we always attempt to tag bug reports with which versions are affected, for example see [this query](https://github.com/aws/aws-for-fluent-bit/issues?q=is%3Aissue+is%3Aopen+label%3Aaws-2.28.0).  
+In the AWS for Fluent Bit repo, we always attempt to tag bug reports with which versions are affected, for example see [this query](https://github.com/aws/aws-for-fluent-bit/issues?q=is%3Aissue+is%3Aopen+label%3Aaws-2.21.0).  
 
 #### Network Connection Issues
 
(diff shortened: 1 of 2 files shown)


troubleshooting/debugging.md
```diff
@@ -191,7 +191,7 @@ If you're running on an old version, it may have been fixed in the latest releas
 
 Or, you may be facing an unresolved or not-yet-discovered issue in a newer version. Downgrading to our latest stable version may work: https://github.com/aws/aws-for-fluent-bit#using-the-stable-tag
 
-In the AWS for Fluent Bit repo, we always attempt to tag bug reports with which versions are affected, for example see [this query](https://github.com/aws/aws-for-fluent-bit/issues?q=is%3Aissue+is%3Aopen+label%3Aaws-2.21.0).
```
PettitWesley, line 194: I would prefer not to remove this. This was an old version on purpose, since we actually have issues tagged for it, so the user can see it work. I do not believe this sentence implies that a user should use 2.21.0. We can edit the wording here instead to make it more clear- like "for example, see this query which shows that 2.21.0 (very old and out of date) has associated issues".

troubleshooting/debugging.md
```diff
@@ -235,10 +235,10 @@ There are some caveats when using Valgrind. Its a debugging tool, and so it sign
...
+Use the Dockerfile below to create a new container image that will invoke your Fluent Bit version using Valgrind. Replace the "builder" image with whatever AWS for Fluent Bit release you are using, in this case, we are testing 2.28.0. The Dockerfile here will copy the original binary into a new image with Valgrind. Valgrind is a little bit like a mini-VM that will run the Fluent Bit binary and inspect the code at runtime. When you terminate the container, Valgrind will output diagnostic information on shutdown, with summary of memory leaks it detected. This output will allow us to determine which part of the code caused the leak. Valgrind can also be used to find the source of segmentation faults
 
 ```
-FROM public.ecr.aws/aws-observability/aws-for-fluent-bit:2.21.0 as builder
+FROM public.ecr.aws/aws-observability/aws-for-fluent-bit:2.28.0 as builder
```
PettitWesley, line 241: I see... I don't think this is a good long term solution since we will need to keep updating it. I'd prefer to put `latest` or `stable` in the example.

PettitWesley, line 241: I see... I don't think this is a good long term solution since we will need to keep updating it. I'd prefer to put `latest` or `stable` in the example.

PettitWesley (changes requested):

troubleshooting/debugging.md
hankwallace: Ok, I'll update it.

troubleshooting/debugging.md
hankwallace: Understood. I'll revert this.

PettitWesley:
@hankwallace Can you also add `--error-limit=no` to the valgrind command in the dockerfile.debug: https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile.debug#L143

PettitWesley (approved):

PettitWesley:
@hankwallace Thanks!

troubleshooting/debugging.md
hankwallace: I went with `latest` since that's the one I suspect people will most likely be debugging.

PR #417, opened 2022-09-07
windows: added the buildspec for creating Windows image manifests

rawahars:
## Summary
Windows images would have the following manifests-
- <AWS_FOR_FLB_VERSION>-windowsservercore
- windowsservercore-latest
- windowsservercore-stable

This change adds the buildspec which would be used in a CodeBuild project to generate the manifests for Windows images.

## Description of changes:
added the buildspec to generate manifests for Windows images

---

### DIFF
--- a/buildspec_windows_create_manifest.yml
+++ b/buildspec_windows_create_manifest.yml
@@ -1,0 +1,53 @@
+version: 0.2
+
+phases:
+  install:
+    commands:
+      - echo "Creating manifest for aws-for-fluent-bit Windows image"
+      - yum update -y && yum upgrade -y
+      - yum install jq -y
+  build:
+    commands:
+      - |
+        export DOCKER_CLI_EXPERIMENTAL=enabled
+        declare -A os_releases=( ["windows2019"]="ltsc2019" ["windows2022"]="ltsc2022")
+        
+        # Push the image to ECR with corresponding architecture as the tag.
+        aws ecr get-login-password --region ${AWS_REGION}| docker login --username AWS --password-stdin ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com
+        
+        # Find the applicable windows versions.
+        cd $CODEBUILD_SRC_DIR
+        versions=$(cat windows.versions | jq -c -r '.windows[].version')
+        latest_version=""
+        
+        # Create and push the image manifests.
+        while read -r version; do
+          if [[ "$latest_version" < "$version" ]]; then
+            latest_version=$version
+          fi
+          
+          docker manifest create ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-windows:${version}-windowsservercore \
+          ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-windows:${version}-${os_releases[windows2019]} \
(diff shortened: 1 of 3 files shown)


buildspec_windows_create_manifest.yml
```diff
@@ -0,0 +1,53 @@
...
+        # Push the image to ECR with corresponding architecture as the tag.
+        aws ecr get-login-password --region ${AWS_REGION}| docker login --username AWS --password-stdin ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com
+        
+        # Find the applicable windows versions.
+        cd $CODEBUILD_SRC_DIR
+        versions=$(cat windows.versions | jq -c -r '.windows[].version')
+        latest_version=""
+        
+        # Create and push the image manifests.
+        while read -r version; do
```
PettitWesley, line 24: should this be `versions`?

PettitWesley (approved):

buildspec_windows_create_manifest.yml
rawahars: These would be the versions defined in `windows.versions` file. The same would have been built in prior stages.

PettitWesley: Cool but the variable that the file is read into above is called `versions` if I am reading this correctly

rawahars: So we will feed all the versions to the loop and in each iteration, we would be reading one version in the variable `version`.

Basically, `versions` would be something like below-
```
2.26.1
2.26.2
```

`read -r version` would read each line in `$version` on each iteration.

rawahars:
Rebased against the latest `aws/windows` branch.

buildspec_windows_create_manifest.yml
```diff
@@ -0,0 +1,53 @@
...
+        
+          # Sanity check the manifest.
+          docker manifest inspect ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-windows:${version}-windowsservercore
+          
+          # Push manifest to ECR.
+          docker manifest push ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-windows:${version}-windowsservercore
+        done <<< "$(echo "$versions")"
+        
+        # Create and push manifest for latest image.
+        docker manifest create ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-windows:windowsservercore-latest \
```
arun-annamalai, line 41: Where do the environment variables such as AWS_ACCOUNT, AWS_REGION get propagated from? The code pipeline upstream or just inherently in the code build project

PettitWesley (approved):

buildspec_windows_create_manifest.yml
rawahars: So these would be pumped in during the infra creation. 
When we would create the CodeBuild projects then these would be either set as environment variables.

integ/run-integ.ps1
```diff
@@ -250,7 +250,7 @@ Function Clean-Test {
     )
 
     # Run "docker-compose down" once the tests have been ran.
-    docker-compose --file $DockerComposeFilePath --profile $CoreProfile --profile $TestProfile down -v --rmi all --remove-orphans
```
jterry75, line 253: Why skip cleanup?

rawahars: Well it was causing some weird errors during the test run wherein few image layers were cleaned up during first test say for `cloudwatch` and then during second test for `cloudwatch_logs`, cleanup+test fails saying that the layers cannot be found.

jterry75 (approved):

PR #430, opened 2022-09-26
windows: OS release specific integration test stacks and configurable repository name in scripts

rawahars:
## Summary

- Currently, a single stack was created for both Windows Server 2019 and Windows Server 2022 OS platforms. This leads to a dependency while running integration tests.
We can run parallel tests for WS2019 and WS2022 if we create different stacks for each OS platform. This change adds the same support.

- Currently, we had hardcoded the repository name where the images and manifests would be pushed to. This leads us to have a dependency on the name of the resources. Instead, we are making the same configurable via a parameter which can be set based on the resource which was created.

## Implementation changes
The scripts already pushed as part of earlier PRs are modified.

## Testing
The code was tested end to end using the branch - https://github.com/rawahars/aws-for-fluent-bit/tree/windows-integ

---

Signed-off: Harsh Rawat <rawahars@amazon.com>


jterry75 (approved):

scripts/create_windows_manifest.sh
```diff
@@ -0,0 +1,58 @@
...
+  docker manifest create ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/${REPOSITORY}:${version}-windowsservercore \
+  ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/${REPOSITORY}:${version}-${os_releases[windows2019]} \
+  ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/${REPOSITORY}:${version}-${os_releases[windows2022]}
+
+  # Sanity check the manifest.
+  docker manifest inspect ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/${REPOSITORY}:${version}-windowsservercore
+
+  # Push manifest to ECR.
+  docker manifest push ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/${REPOSITORY}:${version}-windowsservercore
+done <<< "$(echo "$versions")"
```
PettitWesley, line 43: why this?

rawahars: So while loop in bash is executed inside a subshell. Therefore, any values calculated inside cannot be accessed outside the loop.
To overcome the same, we are using Process Substitution.
Reference: http://mywiki.wooledge.org/BashFAQ/024

PR #443, opened 2022-10-17
Windows: Updated the create manifest script to support multiple registries

rawahars:
## Summary
Presently, we are publishing the manifest of images to `regional ECR`, `Public ECR`, and `DockerHub`. Additionally, we also have a workflow to sync stable image in each region, wherein we create the manifest for the stable image.

Therefore, in this change, we are-
- Creating a common script which has common methods across various Windows scripts which we would be using in the release infrastructure.
    - Methods to set the required variables such as `All supported versions`, `Latest version`, `Stable version`, etc.
    - Method to check if the image manifest for Windows exists in a region.
    - Method to check if the two manifests are same.
- Refactor the existing create manifest script to use multiple registries.
- Changed the CMD in Windows dockerfile to exec format.
- Created a new file for Windows stable version.
- Updated versions in windows.version

## Testing
These scripts have been tested as part of an automated end to end workflow.


scripts/common_windows.sh
```diff
@@ -0,0 +1,130 @@
...
+# language governing permissions and limitations under the License.
+
+set -uo pipefail
+
+#############################################################################################################
+# Helper methods to set the required variables.
+#############################################################################################################
+# Constant
+# Number of presently supported OS versions.
+SUPPORTED_WINDOWS_OS_RELEASES_NUMBER=2
```
jterry75, line 22: What does this mean?

scripts/common_windows.sh
```diff
@@ -0,0 +1,130 @@
...
+    if [[ "$AWS_FOR_FLUENT_BIT_LATEST_VERSION" < "$version" ]]; then
+      AWS_FOR_FLUENT_BIT_LATEST_VERSION=$version
+    fi
+  done <<< "$(echo "$ALL_AWS_FOR_FLUENT_BIT_VERSIONS")"
+}
+
+# Sets AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB variable with the latest version available in dockerhub.
+set_dockerhub_version() {
+  # Find the dockerhub version.
+  docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
```
jterry75, line 49: When we push to dockerhub are they full layers or are they foreign layers?

scripts/common_windows.sh
```diff
@@ -0,0 +1,130 @@
...
+  if [ "${first_manifest}" = 1 ]; then
+    echo "No manifest found for ${first_manifest_image}"
+    return 1
+  elif [ "${second_manifest}" = 1 ]; then
+    echo "No manifest found for ${second_manifest_image}"
+    return 1
+  fi
+
+  # Obtain the digest entries from the manifest.
+  first_manifest_digests=$(echo "${first_manifest}" | jq -r '.manifests[].digest' | sort)
```
jterry75, line 111: This wont work on a manifest list right?

scripts/common_windows.sh
rawahars: So this is essentially the count of number of supported Windows releases. Right now it is 2 i.e. Windows Server 2019 and Windows Server 2022.

In `check_image_manifest_exists` method, we check if the image manifest for the given version does exist in the target registry and the number of digests is equal to this number i.e. 2.

scripts/common_windows.sh
rawahars: So this method does compare manifest lists. `$(echo "${first_manifest}" | jq -r '.manifests[].digest' | sort)` returns a list of digests which we compare for both the images.

scripts/common_windows.sh
rawahars: So as discussed offline, currently we are pushing full layers to dockerhub.

jterry75 (approved):

PettitWesley (approved):

PR #446, opened 2022-10-17
Windows: added the scripts for publishing and verifying SSM Parameters

rawahars:
## Summary
For AWS-for-fluent-bit images, we publish SSM Parameters in each regional store. The normal CodePipeline runs in the leader region of standard partition. As part of that workflow, we publish the SSM Parameters based on new images for each of the classic region in standard partition.

As part of sync-task workflow, we publish the SSM Parameter in each opt-in region as well as China and Gov cloud partition. This sync is based on the updated images in Public ECR repository.

As part of this change, we are adding the following-
- Script to publish and sync SSM parameter in each region
- Script to verify the SSM parameter in each region
- Buildspecs used by CodeBuild projects

## Testing
These scripts have been tested as part of an automated end to end workflow.

### DIFF
--- a/buildspec_windows_verify_ssm_parameters.yml
+++ b/buildspec_windows_verify_ssm_parameters.yml
@@ -1,0 +1,16 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      python: 3.x
+    commands:
+      - echo "Verifies the SSM parameters for Windows"
+      - yum update -y && yum upgrade -y
+      - yum install jq -y
+  build:
+    commands:
+      # Validates that the SSM parameters are published as required.
+      # The required variables would be passed on via CodeBuild Environment variables.
+      - cd ./scripts
+      - './verify_windows_ssm_parameters.sh'
--- a/scripts/publish_windows_ssm_parameter.sh
+++ b/scripts/publish_windows_ssm_parameter.sh
@@ -1,0 +1,201 @@
+#!/bin/bash
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# 	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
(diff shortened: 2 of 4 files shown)


scripts/publish_windows_ssm_parameter.sh
```diff
@@ -0,0 +1,201 @@
...
+# PUBLIC_REGISTRY_ALIAS -> Used when syncing the SSM params.
+# PUBLIC_REPOSITORY -> Public repository which is used for the image sync.
+#############################################################################################################
+
+#############################################################################################################
+# Helper methods to publish the SSM Parameter.
+#############################################################################################################
+
+# This method publishes the SSM parameter for a given tag if the same is not present or obsolete.
+publish_ssm_param_if_obsolete() {
```
PettitWesley, line 32: nit: I feel like obsolete is not the best word here... missing, not present, or backfill are words that to me more clearly fit here

buildspec_windows_verify_ssm_parameters.yml
```diff
@@ -0,0 +1,16 @@
...
+    commands:
+      - echo "Verifies the SSM parameters for Windows"
+      - yum update -y && yum upgrade -y
+      - yum install jq -y
+  build:
+    commands:
+      # Validates that the SSM parameters are published as required.
+      # The required variables would be passed on via CodeBuild Environment variables.
+      - cd ./scripts
+      - './verify_windows_ssm_parameters.sh'
```
PettitWesley, line 16: why separate buildspecs and projects for publishing and verifying?

rawahars: So for the `Build-Images` phase of our pipeline, we did need separate build projects (One for build and one for manifest creation) and consequently, the verify phase was separated. In short, we have the following-
- Build images stage
- Build manifest stage
- Verify image stage

Same goes for `Publish Images` phase as well since we need to create Windows instances to pull and push the images.

To ensure that we had similar semantics across the pipeline, I created SSM stage in two parts as well-
- Publish SSM Parameters
- Verify SSM Parameters

jterry75 (approved):

scripts/publish_windows_ssm_parameter.sh
rawahars: Addressed this comment in the latest revision

PettitWesley (approved):

PR #456, opened 2022-11-08
linux pipeline: auth to dockerhub before calling API

PettitWesley:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -21,10 +21,6 @@ IMAGE_SHA_MATCHED="FALSE"
 AWS_FOR_FLUENT_BIT_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION)
 AWS_FOR_FLUENT_BIT_STABLE_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_STABLE_VERSION)
 
-docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
-tag_array=(`echo ${docker_hub_image_tags}`)
-AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py ${tag_array[@]})
-
 # Enforce STS regional endpoints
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -235,6 +243,8 @@ sync_latest_image() {
 	region=${1}
 	account_id=${2}
 
+	get_latest_dockerhub_version ${region}
```
PettitWesley, line 246: We can't call this function every time this script is used since it will require secrets manager permissions. So now we only call it when needed.

scripts/publish.sh
```diff
@@ -91,6 +87,19 @@ ARCHITECTURES=("amd64" "arm64")
...
+# Sets AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB variable with the latest version available in dockerhub.
+get_latest_dockerhub_version() {
+    username="$(aws secretsmanager get-secret-value --secret-id $DOCKER_HUB_SECRET --region ${1} | jq -r '.SecretString | fromjson.username')"
+    password="$(aws secretsmanager get-secret-value --secret-id $DOCKER_HUB_SECRET --region ${1} | jq -r '.SecretString | fromjson.password')"
+    payload="{\"username\": \"${username}\",\"password\": \"${password}\"}"
+    token=$(curl -s -X POST -H "Content-Type: application/json" -d "${payload}"  https://hub.docker.com/v2/users/login | jq -r '.token')
+    auth="Authorization: Bearer ${token}"
+    docker_hub_image_tags=$(curl -s -S -H "${auth}" 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
+    tag_array=(`echo ${docker_hub_image_tags}`)
+    AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py ${tag_array[@]})
```
matthewfala, line 99: This script doesn't need to be authorized too, does it?

matthewfala: Oh i see. It's just a parser. https://github.com/PettitWesley/aws-for-fluent-bit/blob/auth-to-dockerhub/scripts/get_latest_dockerhub_version.py

PR #458, opened 2022-11-09
debugging: add tail multiple and firehose/kinesis partial success sec…

PettitWesley:
…tions

*Issue #, if available:*

*Description of changes:*


matthewfala (approved):
This looks great!
We should be able to solve the log duplication problem here if we wanted to:
https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html

We would just make a hash map with a key based on the chunk's memory address. The value would be a boolean array of true and false flags based on each records success status. Always check the hash map before sending logs and avoid sending duplicates.

PettitWesley:
@matthewfala agreed- do you want to open an upstream issue for that, since it would be an upstream change

matthewfala:
We do have to consider clean up. We could probably just keep track of the retries and if we're on the last attempt, then get rid of the hash entry.

matthewfala:
Sure!

matthewfala:
https://github.com/fluent/fluent-bit/issues/6369

PR #465, opened 2022-11-10
Update gdb debugging

ShelbyZ:
Signed-off-by: Shelby Hagman [shagman@amazon.com](mailto:shagman@amazon.com)

Issue #, if available:

Description of changes:
- Add yum-utils to Dockerfile.debug for debuginfo-install tool
- Add section about gdb listing missing debuginfos and example

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -118,6 +118,7 @@ RUN yum upgrade -y \
           valgrind \
           libyaml \
           gdb \
+          yum-utils \
           nc && rm -fr /var/cache/yum
 
 COPY --from=builder /fluent-bit /fluent-bit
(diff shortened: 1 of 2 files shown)


Dockerfile.debug
```diff
@@ -118,6 +118,7 @@ RUN yum upgrade -y \
           valgrind \
           libyaml \
           gdb \
+          yum-utils \
```
PettitWesley, line 121: This is necessary to get debuginfo-install?

Do we also need to remove the next line where it removes yum cache or not?

This has been tested right or do I need to pull it and check it myself?

ShelbyZ: This is one way to get debuginfo-install - https://linux.die.net/man/1/yum-utils, we could optionally try to install [python3-debuginfo](https://serverfault.com/questions/486764/how-to-install-python-debug-symbols-for-gdb-on-amazon-linux-ami) but it is more mess.

We leave the line after to handle cache cleanup. Building the image via `make debug` was done without issue and I ran a local image to check that `debuginfo-install` was available.

PettitWesley: Thanks Shelby!

PettitWesley (approved):

PR #467, opened 2022-11-11
debugging: add FireLens tag and match pattern

PettitWesley:

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -40,6 +40,7 @@
 - [Fluent Bit Windows containers](#fluent-bit-windows-containers)
     - [Networking issue with Windows containers when using async DNS resolution by plugins](#networking-issue-with-windows-containers-when-using-async-dns-resolution-by-plugins)
 - [FAQ]
+    - [FireLens Tag and Match Pattern and generated config](#firelens-tag-and-match-pattern-and-generated-config)
 
 
 ### Understanding Error Messages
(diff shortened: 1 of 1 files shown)


troubleshooting/debugging.md
```diff
@@ -479,3 +480,30 @@ Path                /logs/request.log
...
+So for example, if your container name is `web-app` and then your Task ID is `4444-4444-4444-4444` then the tag assigned to stdout/stderr logs from the `web-app` container will be:
+
+```
+web-app-firelens-4444-4444-4444-4444
+```
+
+If you use a custom config file with FireLens, then you would want to set a `Match` pattern for these logs like:
+
+```
+Match web-app-firelens*
```
coilysiren, line 506: Hi again! I was just looking through your repo for an example of this exact thing, and your PR here came up in my search! Very fun timing.

My specific use case here is wanting to match on everything coming out of firelens, so I can send it to an error monitoring service. I imagine this more common use case than wanting to match on a specific container. So my suggestion is that this section have something that reflects that, like so:

```
Match *firelens*
```

```
[FILTER]
    Name rewrite_tag
    Match *firelens*
    Rule $level ^(error|ERROR) errors true
```

Just my 2 cents - I've since figured out my own configuration thanks to all your other docs. Thanks so much!

PettitWesley: Good idea, I'll add that :)

PettitWesley: That helps makes it clear that Match is not just for prefix matching

matthewfala (approved):

DrewZhang13 (approved):

PR #468, opened 2022-11-11
windows: consolidate all configurable versions in the config file

rawahars:
## Summary
Presently, we have OpenSSL and FlexBison versions hardcoded in the build scripts itself. This means that changing the same requires a change to the build script. Additionally, the versions are then scattered across all the build files making it harder to understand.

Therefore, as part of this change, we are consolidating all in the master config file for Windows.

## Description of changes:
windows: consolidate all configurable versions in the config file

### DIFF
--- a/windows.versions
+++ b/windows.versions
@@ -6,7 +6,9 @@
       "fluent-bit": "1.9.9",
       "kinesis-plugin": "v1.10.0",
       "firehose-plugin": "v1.7.0",
-      "cloudwatch-plugin": "v1.9.0"
+      "cloudwatch-plugin": "v1.9.0",
+      "openssl": "3.0.7",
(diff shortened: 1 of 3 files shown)


windows.versions
```diff
@@ -8,5 +8,7 @@
       "firehose-plugin": "v1.7.0",
       "cloudwatch-plugin": "v1.9.0"
     }
-  ]
+  ],
+  "openssl": "3.0.7",
+  "flexBison": "2.5.22"
```
PettitWesley, line 13: why isn't it part of a specific release in the list like the others??

PettitWesley (approved):

PettitWesley (approved):

PR #499, opened 2022-12-15
WIP: ECS Fluent Bit Daemon Collector

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ecs/daemon/inputs/ecs/agent.conf
+++ b/ecs/daemon/inputs/ecs/agent.conf
@@ -1,8 +1,8 @@
 [INPUT]
     Name                tail
     Tag                 ecs.*
-    Path                /var/log/ecs/*
-    DB                  /var/fluent-bit/state/flb_ecs.db
+    Path                /var/log/ecs/ecs-agent.log*
+    DB                  /var/fluent-bit/state/flb_ecs_agent.db
     Mem_Buf_Limit       50MB
     Skip_Long_Lines     On
     Refresh_Interval    10
--- a/ecs/daemon/inputs/ecs/agent.conf
+++ b/ecs/daemon/inputs/ecs/agent.conf
@@ -1,12 +1,11 @@
 [INPUT]
     Name                tail
     Tag                 ecs.*
-    Path                /var/log/ecs/ecs-agent.log*
+    Path                /var/log/ecs/ecs-agent.log
     DB                  /var/fluent-bit/state/flb_ecs_agent.db
     Mem_Buf_Limit       50MB
     Skip_Long_Lines     On
     Refresh_Interval    10
     Rotate_Wait         30
     storage.type        filesystem
-    # Collect all logs on instance: On vs Off
     Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
--- a/ecs/daemon/inputs/ecs/agent.conf
+++ b/ecs/daemon/inputs/ecs/agent.conf
@@ -1,0 +1,12 @@
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/*
+    DB                  /var/fluent-bit/state/flb_ecs.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    # Collect all logs on instance: On vs Off
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
--- a/ecs/daemon/inputs/ecs/all-ecs-dataplane.conf
+++ b/ecs/daemon/inputs/ecs/all-ecs-dataplane.conf
@@ -1,12 +1,86 @@
+# It'd be nice if this could be one tail input reading `/var/log/ecs/*`
+# However that will match rotated files: https://github.com/aws/aws-for-fluent-bit/blob/mainline/troubleshooting/debugging.md#tail-input-duplicates-logs-because-of-rotation
+
 [INPUT]
     Name                tail
     Tag                 ecs.*
-    Path                /var/log/ecs/*
-    DB                  /var/fluent-bit/state/flb_ecs.db
+    Path                /var/log/ecs/ecs-agent.log
+    DB                  /var/fluent-bit/state/flb_ecs_agent.db
     Mem_Buf_Limit       50MB
     Skip_Long_Lines     On
     Refresh_Interval    10
     Rotate_Wait         30
     storage.type        filesystem
-    # Collect all logs on instance: On vs Off
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
+
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/audit.log
+    DB                  /var/fluent-bit/state/flb_ecs_audit.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
+
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/ecs-cni-bridge-plugin.log
+    DB                  /var/fluent-bit/state/flb_ecs_cni_bridge.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
+
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/ecs-cni-eni-plugin.log
+    DB                  /var/fluent-bit/state/flb_ecs_cni_eni.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
+
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/ecs-init.log
+    DB                  /var/fluent-bit/state/flb_ecs_init.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
+
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/ecs-volume-plugin.log
+    DB                  /var/fluent-bit/state/flb_ecs_volume.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
+
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/vpc-branch-eni.log
+    DB                  /var/fluent-bit/state/flb_ecs_branch_eni.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
     Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
--- a/ecs/daemon/inputs/ecs/all-ecs-dataplane.conf
+++ b/ecs/daemon/inputs/ecs/all-ecs-dataplane.conf
@@ -1,0 +1,12 @@
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/*
+    DB                  /var/fluent-bit/state/flb_ecs.db
+    Mem_Buf_Limit       50MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
+    # Collect all logs on instance: On vs Off
+    Read_From_Head      ${READ_FROM_HEAD_ECS_DATAPLANE}
--- a/ecs/daemon/inputs/host/all-host.conf
+++ b/ecs/daemon/inputs/host/all-host.conf
@@ -3,6 +3,7 @@
     # Tag is set with underscores for cloud init because s3 output option
     # s3_key_format_tag_delimiters includes dash
     Tag                 host.var.log.cloud_init
+    # Cloud init runs once on startup, so we want to collect all rotated files
     Path                /var/log/cloud-init.log*
     DB                  /var/fluent-bit/state/flb_cloud_init.db
     Mem_Buf_Limit       5MB
@@ -19,6 +20,7 @@
     # Tag is set with underscores for cloud init because S3 output option
     # s3_key_format_tag_delimiters includes dash
     Tag                 host.var.log.cloud_init_output
+    # Cloud init runs once on startup, so we want to collect all rotated files
     Path                /var/log/cloud-init-output.log*
     DB                  /var/fluent-bit/state/flb_cloud_init_output.db
     Mem_Buf_Limit       5MB
@@ -34,7 +36,7 @@
 [INPUT]
     Name                tail
     Tag                 host.*
-    Path                /var/log/secure*
+    Path                /var/log/secure
     Parser              syslog
     DB                  /var/fluent-bit/state/flb_secure.db
     Mem_Buf_Limit       5MB
@@ -45,7 +47,7 @@
 [INPUT]
     Name                tail
     Tag                 host.*
-    Path                /var/log/messages*
+    Path                /var/log/messages
     Parser              syslog
     DB                  /var/fluent-bit/state/flb_messages.db
     Mem_Buf_Limit       5MB
@@ -56,7 +58,7 @@
 [INPUT]
     Name                tail
     Tag                 host.*
-    Path                /var/log/dmesg*
+    Path                /var/log/dmesg
     Parser              syslog
     DB                  /var/fluent-bit/state/flb_dmesg.db
     Mem_Buf_Limit       5MB
--- a/ecs/daemon/inputs/host/all-host.conf
+++ b/ecs/daemon/inputs/host/all-host.conf
@@ -1,0 +1,65 @@
+[INPUT]
+    Name                tail
+    # Tag is set with underscores for cloud init because s3 output option
+    # s3_key_format_tag_delimiters includes dash
+    Tag                 host.var.log.cloud_init
+    Path                /var/log/cloud-init.log*
+    DB                  /var/fluent-bit/state/flb_cloud_init.db
+    Mem_Buf_Limit       5MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Rotate_Wait         30
+    storage.type        filesystem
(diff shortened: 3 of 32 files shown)


ecs/daemon/outputs/s3/task.conf
```diff
@@ -0,0 +1,45 @@
...
+
+
+[OUTPUT]
+    Name                         s3
+    Match                        task.*
+    bucket                       ${BUCKET_TASK}
+    region                       ${REGION_TASK}
+    upload_timeout               2m
+    total_file_size              10M
+    s3_key_format                /aws/ecs/${CLUSTER_NAME}/$TAG[0]/$TAG[1]/$TAG[2]/$TAG[3]/$TAG[4]/%Y/%m/%d/%H/%M/%S
```
PettitWesley, line 33: Matt: do we really want pseudo directories with `/` in the timestamp. May be only do it do day, not below that.

ecs/daemon/outputs/s3/task.conf
```diff
@@ -0,0 +1,45 @@
...
+    store_dir                    /var/fluent-bit/s3-store/task
+
+[OUTPUT]
+    Name                         s3
+    Match                        app.*
+    bucket                       ${BUCKET_TASK}
+    region                       ${REGION_TASK}
+    upload_timeout               2m
+    total_file_size              10M
+    s3_key_format                /aws/ecs/${CLUSTER_NAME}/task/failed-to-retrieve-metadata-for/$TAG/%Y/%m/%d/%H/%M/%S
```
PettitWesley, line 44: If I have two rewrite_tag filters, one for the fallback logs, I could still add instance ID probably to them, to make it still somewhat possible to figure where the failed logs came from.

ecs/daemon/outputs/cloudwatch_logs/task.conf
```diff
@@ -0,0 +1,30 @@
...
+    Name                   cloudwatch_logs
+    Match                  task.var.lib.docker.containers.*
+    # log group/stream names with metadata templating
+    log_group_template     /aws/ecs/${CLUSTER_NAME}/application/$ecs_family
+    log_stream_template    $ec2_instance_id.$ecs_task_id.$container
+    # fallback group if templating fails
+    log_group_name         /aws/ecs/${CLUSTER_NAME}/application/fallback-group-for-failed-metadata-templating
+    # fallback stream to log_stream_prefix + tag (log file name) if templating fails
+    log_stream_prefix      json-log-driver-file-
+    region                 ${REGION_TASK}
```
PettitWesley, line 28: bingfeng: So this enables cross region log, but what if customers want cross account log? 

wesley: That would require a different output file.

PettitWesley: cross account requires you to add a new parameter in teh config:

```
log_stream_prefix      json-log-driver-file-
region                 ${REGION_TASK}
role_arn              ${ROLE_ARN_TASK}
```

which then can't be empty or it will fail validation.

PettitWesley: we can use the EKS helm chart as an inspiration:

```
log_stream_prefix      json-log-driver-file-
region                 ${REGION_TASK}
${EXTRA_CLOUDWATCH_TASK}
```

its safe for `EXTRA_CLOUDWATCH_TASK`

Need to check if this really works. 

set `EXTRA_CLOUDWATCH_TASK=`role_arn   my-role`

PettitWesley: need to think about the experience

PettitWesley:
```
log_stream_prefix      json-log-driver-file-
region                 ${REGION_TASK}
${OPTIONAL_ROLE_ARN}
${OPTIONAL_LOG_KEY}
${OPTIONAL_LOG_RETENTION}
```

all of these **I THINK** but need to check, can be empty and it won't break.

PettitWesley: If this doesn't work, then may be need an init feature which is like what the helm chart.

This might add a newline each time if its empty which won't work.

PettitWesley: need to check.

PettitWesley: basically init templating of config

matthewfala: ## Templating Documentation
One option discussed in today's meeting which I can put here for documentation is as follows:
Have a single environment variable called "init-config" which can be set to a JSON string. JSON is great because any JSON can be reduced to a one line string, thus one environment variable.


May take some more thinking to settle on a good JSON schema but the following would work
#### Configuration JSON
```
{
    /* These are global configuration key value pairs */
    "config": {
        "param1": "val"
    },
    "options": [
         /* Here are a list of options that the customer can configure separately. Name describes the option they are opting into, config is the option's setting */
         {
            "name": "host",
            /* Each config has a default setting, so config is completely optional */
            "config": {
                "destinations": ["s3", "cloudwatch"]
            }
         },
         {
            "name": "host",
            "config": {
                "destination": "cloudwatch",
                "region": "auto" /* Each config gets translated into a template model which gets translated to the template. Non trivial config options should be considered */
            }
         }
    ]
}
```


### Fluent Bit Config Generators
Instead of having init process stitch together raw config files, we introduce a concept of data models and config file templates.

A process on startup parses the JSON config and generates a set of configuration files which are used to startup fluent bit (similar to the init process but with an added step of generating the config files from template)

**Templates**
We can use a widely used templating engine to generate our fluent bit config files adaptively from our config file.
I recommend Mustache render https://mustache.github.io/
It's great, super flexible and allows for conditional and looped blocks of template content which will no doubt come in handy for some future features. See Demo: https://mustache.github.io/#demo

**Data Models**
A mustache template needs a data model to render into it's result. Datamodels are just object. We can generate a Data Model for each template.

**Generator Structure**
This could be how we store the template rendering logic and organize our generator scripts
```
{
    "options": [
        /* template providers for the options we can enable */
        {
            "name": "<OptionName>" /* this option name corresponds to the options that can be enabled */
            "synthesizeDataModel": function(globalConfig, optionConfig) => someObject, /* a function that makes our template data model */
            "templates": [
                /* A list of templates that get processed with our data model and sent to the output destination conf file */
                {
                    "path": "/anyPathToOurTemplate",
                    "destination": "someDestinationConfigFile" /* append to destination file, maybe /parser.conf or /fluent-bit.conf */
                },
               ...
            ]
        }
    ]
}
```

(I highly recommend Typescript, because I love TS and could code this efficiently, but adding node to our image could introduce security scan vulnerabilities that we may have to patch which is unfortunate)

Note: I actually don't like the proposed schemas and structures. It's just a general idea of some thoughts (initial thoughts).

ecs/daemon/inputs/host/all-host.conf
```diff
@@ -0,0 +1,65 @@
+[INPUT]
```
PettitWesley, line 1: I keep forgetting my own guidance here: https://github.com/aws/aws-for-fluent-bit/blob/mainline/troubleshooting/debugging.md#tail-input-duplicates-logs-because-of-rotation

I need to double check and consider all of these. For each, there's a trade-off:
1. If we match rotated log files, then we will get duplicates
2. If we do not match all of the existing rotated log files at the time when Fluent Bit starts, then we won't get all of those old logs. Probably this is okay, but it'd be cool if there was a solution for it to get all the old logs. The ECS log collector script will grab all old logs IIRC: https://github.com/aws/amazon-ecs-logs-collector

ecs/daemon/inputs/ecs/all-ecs-dataplane.conf
```diff
@@ -0,0 +1,12 @@
+[INPUT]
```
PettitWesley, line 1: I keep forgetting my own guidance here: https://github.com/aws/aws-for-fluent-bit/blob/mainline/troubleshooting/debugging.md#tail-input-duplicates-logs-because-of-rotation

I need to double check and consider all of these. For each, there's a trade-off:
1. If we match rotated log files, then we will get duplicates
2. If we do not match all of the existing rotated log files at the time when Fluent Bit starts, then we won't get all of those old logs. Probably this is okay, but it'd be cool if there was a solution for it to get all the old logs. The ECS log collector script will grab all old logs IIRC: https://github.com/aws/amazon-ecs-logs-collector

ecs/daemon/inputs/host/all-host.conf
```diff
@@ -0,0 +1,65 @@
...
+    Mem_Buf_Limit       5MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Read_from_Head      ${READ_FROM_HEAD_HOST}
+
+[INPUT]
+    Name                tail
+    Tag                 host.*
+    Path                /var/log/dmesg*
+    Parser              syslog
```
PettitWesley, line 60: My comment here is applicable: https://github.com/aws-samples/amazon-cloudwatch-container-insights/pull/108#issuecomment-1353804758

ecs/daemon/inputs/task.conf
```diff
@@ -0,0 +1,16 @@
+[INPUT]
+    Name                tail
+    Tag                 app.*
+    Path                /var/lib/docker/containers/*/*.log
+    Docker_Mode         On
+    Docker_Mode_Flush   5
+    Docker_Mode_Parser  container_firstline
+    Parser              docker
```
PettitWesley, line 8: As I note here, this is the old parser config, we can/should use the new one which supports docker: https://github.com/aws-samples/amazon-cloudwatch-container-insights/pull/115#discussion_r1062927891

ecs/daemon/inputs/ecs/agent.conf
```diff
@@ -0,0 +1,12 @@
+[INPUT]
+    Name                tail
+    Tag                 ecs.*
+    Path                /var/log/ecs/ecs-agent.log*
```
PettitWesley, line 4: As noted elsewhere, need to not match rotated agent log files

ecs/daemon/inputs/ecs/all-ecs-dataplane.conf
PettitWesley: I think this is something to just note in the debugging/DIY part of the guide

ecs/daemon/inputs/host/all-host.conf
PettitWesley: fixed/will be noted in DIY part of the guide

ecs/daemon/inputs/host/all-host.conf
PettitWesley: I think for this daemon service template, putting the dmesg log inside of a key called "log" is all we want... so just removing the parser here.

ecs/daemon/inputs/host/all-host.conf
```diff
@@ -0,0 +1,67 @@
...
+    Mem_Buf_Limit       5MB
+    Skip_Long_Lines     On
+    Refresh_Interval    10
+    Read_from_Head      ${READ_FROM_HEAD_HOST}
+
+[INPUT]
+    Name                tail
+    Tag                 host.*
+    Path                /var/log/messages
+    Parser              syslog
```
PettitWesley, line 51: TODO: Check if this parser is correct and actually most ideal for all of these logs

https://github.com/aws-samples/amazon-cloudwatch-container-insights/pull/108#issuecomment-1353804758

ecs/daemon/parsers.conf
```diff
@@ -0,0 +1,23 @@
+[PARSER]
```
PettitWesley, line 1: TODO: I think we can just use the built-in docker parser. Only syslog may be needed here. Testing this.

use_cases/ecs-ec2-daemon-collector/task-definition-templates/cloudwatch-task-logs.json
```diff
@@ -0,0 +1,79 @@
...
+    ],
+    "containerDefinitions": [
+        {
+            "essential": true,
+            "image": "public.ecr.aws/aws-observability/aws-for-fluent-bit:init-stable",
+            "name": "fluent-bit-daemon-collector",
+            "logConfiguration": {
+                "logDriver": "awslogs",
+                "options": {
+                    "awslogs-group": "/aws/ecs/{INSERT CLUSTER_NAME}/collector",
```
PettitWesley, line 28: should add cache-disabled true here: https://docs.docker.com/config/containers/logging/dual-logging/

PettitWesley: and in all other templates

(Assuming ECS accepts this option and doesn't validate it against a set of options that is old and doesn't include it...)

use_cases/ecs-ec2-daemon-collector/task-definition-templates/cloudwatch-all-logs.json
```diff
@@ -0,0 +1,149 @@
...
+                },
+                {
+                    "sourceVolume": "host-and-ecs-logs",
+                    "containerPath": "/var/log",
+                    "readOnly": true
+                },
+                {
+                    "sourceVolume": "fluent-bit-buffer-directory",
+                    "containerPath": "/var/fluent-bit",
+                    "readOnly": true
```
PettitWesley, line 144: Need to fix this everywhere, the buffer and state dirs should not be readonly.

ecs/daemon/outputs/cloudwatch_logs/task.conf
PettitWesley: Just having the log_key option, either via this or just another config file, for the task logs, is something I kind of want to have even just for the first launch 🤔

PettitWesley: I think doing the `log_key log` thing is what a lot of users want for their container logs. They don't want them as JSONs.

use_cases/ecs-ec2-daemon-collector/task-definition-templates/cloudwatch-task-logs.json
```diff
@@ -0,0 +1,83 @@
...
+                    "awslogs-region": "{INSERT REGION}",
+                    "awslogs-create-group": "true",
+                    "awslogs-stream-prefix": "fluent-bit-stdout-stderr-"
```
PettitWesley, line 31: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html

prefix-name/container-name/ecs-task-id


So I need to remove the dash here

PettitWesley: and may be just make he prefix be `stdout-stderr` since container name is `fluent-bit-daemon-collector`

use_cases/ecs-ec2-daemon-collector/task-definition-templates/cloudwatch-task-logs.json
```diff
@@ -0,0 +1,83 @@
...
+        {
+            "essential": true,
+            "image": "public.ecr.aws/aws-observability/aws-for-fluent-bit:init-stable",
+            "name": "fluent-bit-daemon-collector",
+            "logConfiguration": {
+                "logDriver": "awslogs",
+                "options": {
+                    "awslogs-group": "/aws/ecs/{INSERT CLUSTER_NAME}/collector",
+                    "awslogs-region": "{INSERT REGION}",
+                    "awslogs-create-group": "true",
```
PettitWesley, line 30: I should remove this/set it to false. Log Groups are normally managed by infra as code outside of tasks so that they can be easily cleaned up. 

Also, the [AmazonECSTaskExecutionRolePolicy](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html) does not have create group permissions so if I include this then I require customers to have more perms on their execution or instance role.

Instead, log group can be created with AWS CLI or in the CFN template.

use_cases/ecs-ec2-daemon-collector/task-definition-templates/cloudwatch-task-logs.json
```diff
@@ -0,0 +1,83 @@
...
+                    "awslogs-stream-prefix": "fluent-bit-stdout-stderr-"
+                }
+            },
+            "environment": [
```
PettitWesley, line 34: Need to add an env var here for IMDS version used by AWS Filter.

ecs/daemon/outputs/cloudwatch_logs/task.conf
```diff
@@ -0,0 +1,30 @@
...
+    Match             task.var.lib.docker.containers.*
+    ECS_Tag_Prefix    task.var.lib.docker.containers.
+    ADD               container    $ECSContainerName
+    ADD               ecs_task_id  $TaskID
+    ADD               ecs_family   $TaskDefinitionFamily
+
+[FILTER]
+    Name               aws
+    Match              task.var.lib.docker.containers.*
+    imds_version       v1
```
PettitWesley, line 13: In all config files, IMDS version used by AWS filters should be configured by the same single env var setting which defaults to v1.

ecs/daemon/outputs/s3/task.conf
PettitWesley: Need to update timestamp here as well

ecs/daemon/outputs/cloudwatch_logs/task.conf
PettitWesley, line 28: I think these should be LOG_REGION_X or all just LOG_REGION

ecs/daemon/inputs/task.conf
```diff
@@ -0,0 +1,13 @@
+[INPUT]
+    Name                tail
+    Tag                 app.*
+    Path                /var/lib/docker/containers/*/*.log
```
PettitWesley, line 4: I've tested that if I add:

```
    Exclude_Path        *container-cached*
```

This will ignore the container cache log files, so we only pick up containers that explicitly configure json-file log driver.

PettitWesley: Except, I saw on my instance that the ECS Agent seems to somehow have non-cache log files, I am not sure how this happened:

Screen Shot 2023-01-21 at 9 28 18 PM

PettitWesley: This shows that it does indeed work for real cache files: see the excluded debug message: 
Screen Shot 2023-01-21 at 9 33 27 PM

PettitWesley: I did a simplified test where I scanned one container sub-directory on my test instance ^

PettitWesley: I checked the agent thing on another instance, it seems that the agent typically has json file log driver files. This is strange and unexpected. 

I should look into this/ask ECS Agent team if this is on purpose. I don't think it is.

ecs/daemon/outputs/cloudwatch_logs/ecs.conf
```diff
@@ -0,0 +1,19 @@
+[FILTER]
+    Name               aws
+    Match              ecs.*
+    imds_version       v1
```
PettitWesley, line 4: TODO: IMDS version needs to be an env var everywhere.

use_cases/ecs-ec2-daemon-collector/task-definition-templates/cloudwatch-task-logs.json
PettitWesley: Actually, since we use host mode, its always safe to use V2 with default hop limit of one, so will just use that. No need to worry about this. V2 is always enabled by default now.

ecs/daemon/outputs/cloudwatch_logs/task.conf
PettitWesley: Actually, since we use host mode, its always safe to use V2 with default hop limit of one, so will just use that. No need to worry about this. V2 is always enabled by default now.

ecs/daemon/outputs/cloudwatch_logs/ecs.conf
PettitWesley: Actually, since we use host mode, its always safe to use V2 with default hop limit of one, so will just use that. No need to worry about this. V2 is always enabled by default now.

ecs/daemon/inputs/task.conf
PettitWesley: https://github.com/aws/amazon-ecs-agent/issues/3546

PR #500, opened 2022-12-15
debugging: add debug-init make directive

matthewfala:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -147,12 +147,12 @@ EXPOSE 2020
 # Default: Run under valgrind
 # alternatively, you can change this to run FB directly and try
 # to get a core file instead
-CMD valgrind --leak-check=full --error-limit=no /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
+# CMD valgrind --leak-check=full --error-limit=no /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
 
 # Alternate: Run directly and get core file
 # make sure you run with `docker run --ulimit core=-1`
 # and mount the /cores folder onto the host
 # Un-Comment the lines below:
 # 
-# RUN mkdir /cores && chmod 777 /cores
-# CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
+RUN mkdir /cores && chmod 777 /cores
+CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -156,4 +156,3 @@ CMD valgrind --leak-check=full --error-limit=no /fluent-bit/bin/fluent-bit -c /f
 # 
 # RUN mkdir /cores && chmod 777 /cores
 # CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
- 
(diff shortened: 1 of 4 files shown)


Dockerfile.debug-init
```diff
@@ -0,0 +1,37 @@
...
+RUN go env -w GOPROXY=direct
+
+# Build init process for Fluent Bit
+COPY /init/fluent_bit_init_process.go /
+COPY /go.mod /
+COPY /go.sum /
+RUN go mod tidy \
+    && go build fluent_bit_init_process.go
+
+FROM amazon/aws-for-fluent-bit:debug
```
matthewfala, line 27: This is the only different line

Dockerfile.debug
```diff
@@ -156,4 +156,3 @@ CMD valgrind --leak-check=full --error-limit=no /fluent-bit/bin/fluent-bit -c /f
 # 
 # RUN mkdir /cores && chmod 777 /cores
 # CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
-
```
matthewfala, line 159: I don't know why the diff includes this line.

PettitWesley: your editor doesn't like whitespace at end of file may be?

Dockerfile.debug-init
PettitWesley: @matthewfala I started working on this a while ago but stopped because it was a bit more complicated and time consuming than I had hoped. See the init controls the fluent bit command, so if we truly want a debug dockerfile with a core dump and a valgrind option we need a change in the init process as well. 

I guess its fine not to do that right now... having core dump debug init build is a good first step. I wonder if we can re-use the same dockerfile if this is the only line changed- could we use docker args to determine which is the base image? Similar to what we have in our Dockerfile.plugins: https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile.plugins#L21

PettitWesley: Here was the previous work I tried to do: https://github.com/PettitWesley/aws-for-fluent-bit/commit/1c1e52e82d789d2879d2288baa8d72167079271b

Makefile
```diff
@@ -55,6 +55,12 @@ debug:
 	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
 	docker build --no-cache -t amazon/aws-for-fluent-bit:debug -f Dockerfile.debug .
 
+.PHONY: debug-init
+debug:
+	docker build --no-cache -t aws-fluent-bit-plugins:latest -f Dockerfile.plugins .
```
PettitWesley, line 60: @matthewfala can you also update our debug targets for the step where we build the linux go plugins. That no longer works simply with the Dockerfile.plugins like we see here, you need to make the make debug targets depend on the `linux-plugins` target. Same as we do for `release` target.

Dockerfile.debug
```diff
@@ -147,13 +147,12 @@ EXPOSE 2020
...
 # Alternate: Run directly and get core file
 # make sure you run with `docker run --ulimit core=-1`
 # and mount the /cores folder onto the host
 # Un-Comment the lines below:
 # 
-# RUN mkdir /cores && chmod 777 /cores
-# CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
- 
+RUN mkdir /cores && chmod 777 /cores
+CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf
```
PettitWesley, line 158: Why change the default for the debug dockerfile from valgrind to core option?

matthewfala: You're right. I don't like this.

Makefile
matthewfala: hmm yep.

matthewfala:
Thank you, Wesley. Made appropriate changes in the reorganized Dockerfile pr: Please see: https://github.com/aws/aws-for-fluent-bit/pull/548

PettitWesley:
in-active PRs should just be closed

PR #524, opened 2023-01-26
windows: add debugging option to the entrypoint script

rawahars:
## Summary
In order to obtain code dumps on Windows, we need to set a registry key which specifies the dump folder as well as the level of dump information. This is as per the Microsoft documentation- https://learn.microsoft.com/en-us/windows/win32/wer/collecting-user-mode-dumps

We will be adding a parameter in the entrypoint.ps1 script which the customers can enable to generate the core dumps. Since these core dump files would be generated inside the container, we need to bind mount the location to the host to retrieve the same.

## Testing
Created custom Windows images based on the new configuration and tested the same.

## Description of changes:
windows: add debugging option to the entrypoint script

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -43,6 +43,7 @@
     - [Always use multiline in the tail input](#always-use-multiline-in-the-tail-input)
     - [Tail Input duplicates logs during rotation](#tail-input-duplicates-logs-because-of-rotation)
 - [Fluent Bit Windows containers](#fluent-bit-windows-containers)
+    - [Enabling debug mode for Fluent Bit Windows images](#enabling-debug-mode-for-fluent-bit-windows-images)
     - [Networking issue with Windows containers when using async DNS resolution by plugins](#networking-issue-with-windows-containers-when-using-async-dns-resolution-by-plugins)
 - [Testing](#testing)
     - [Simple TCP Logger Script](#simple-tcp-logger-script)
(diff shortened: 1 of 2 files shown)


troubleshooting/debugging.md
```diff
@@ -608,6 +609,23 @@ In other cases, the original config would be correct. For example, assume your a
...
 ### Fluent Bit Windows containers
 
+#### Enabling debug mode for Fluent Bit Windows images
+Debug mode can be enabled for Fluent Bit Windows images so that when Fluent Bit crashes, we are able to generate crash dumps which would help in debugging the issue. This can be accomplished by overriding the entrypoint command to include the `-EnableCoreDump` flag. For example-
+```
+# If you are using the config file located at C:\fluent-bit\etc\fluent-bit.conf 
+Powershell -Command C:\\entrypoint.ps1 -EnableCoreDump
+
+# If you are using a custom config file located at other location say C:\fluent-bit-config\fluent-bit.conf. 
+Powershell -Command C:\\entrypoint.ps1 -EnableCoreDump -ConfigFile C:\\fluent-bit-config\\fluent-bit.conf
```
PettitWesley, line 619: just curious why double backslash?

troubleshooting/debugging.md
```diff
@@ -608,6 +609,23 @@ In other cases, the original config would be correct. For example, assume your a
 
 ### Fluent Bit Windows containers
 
+#### Enabling debug mode for Fluent Bit Windows images
+Debug mode can be enabled for Fluent Bit Windows images so that when Fluent Bit crashes, we are able to generate crash dumps which would help in debugging the issue. This can be accomplished by overriding the entrypoint command to include the `-EnableCoreDump` flag. For example-
+```
+# If you are using the config file located at C:\fluent-bit\etc\fluent-bit.conf 
+Powershell -Command C:\\entrypoint.ps1 -EnableCoreDump
```
PettitWesley, line 616: what is `C:\\entrypoint.ps1` - where do I find it? 

I don't see it top level in our repo like the linux entrypoint. Sorry this is probably something I should know which you can help me learn now.

rawahars: This script is present in our repository at [this location](https://github.com/aws/aws-for-fluent-bit/blob/mainline/scripts/entrypoint.ps1). 

We are [copying the same](https://github.com/aws/aws-for-fluent-bit/blob/7dba31fa1015966ae0749c9811046fc6b9f69831/Dockerfile.windows#L22) into the root of our container image.

troubleshooting/debugging.md
rawahars: It is just escaping the backslashes. These may not be necessary but since Windows treats double backslashes as singles ones, I put them there.

troubleshooting/debugging.md
```diff
@@ -608,6 +609,23 @@ In other cases, the original config would be correct. For example, assume your a
...
+
+# If you are using a custom config file located at other location say C:\fluent-bit-config\fluent-bit.conf. 
+Powershell -Command C:\\entrypoint.ps1 -EnableCoreDump -ConfigFile C:\\fluent-bit-config\\fluent-bit.conf
+```
+
+The crash dumps would be generated at location `C:\fluent-bit\CrashDumps` inside the container. Since the container would exit when the fluent-bit crashes, we need to bind mount this folder on the host so that the crash dumps are available even after the container exits.
+
+###### Caveat with debug mode 
+Please note that when you are running the Windows images in debug mode, then Golang plugins would not be supported. This means that even though you can use `cloudwatch`, `kinesis`, and `firehose` in normal mode, the same would not be available in debug mode.
```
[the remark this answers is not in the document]
rawahars: I have changed the wordings in the guide.

troubleshooting/debugging.md
PettitWesley: Ah thanks

troubleshooting/debugging.md
```diff
@@ -608,6 +609,25 @@ In other cases, the original config would be correct. For example, assume your a
...
+
+# If you are using a custom config file located at other location say C:\fluent-bit-config\fluent-bit.conf. 
+Powershell -Command C:\\entrypoint.ps1 -EnableCoreDump -ConfigFile C:\\fluent-bit-config\\fluent-bit.conf
+```
+
+The crash dumps would be generated at location `C:\fluent-bit\CrashDumps` inside the container. Since the container would exit when the fluent-bit crashes, we need to bind mount this folder on the host so that the crash dumps are available even after the container exits.
+
+**Note:** This functionality is available in all `aws-for-fluent-bit` Windows images with version `2.31.0` and above. 
+
+###### Caveat with debug mode
```
PettitWesley, line 626: Nit: i think next level indent would be 5 hashes, since previous was 4

PettitWesley (approved):

PR #533, opened 2023-02-01
runbook: add FireLens crash runbook and template, `make core` build target

PettitWesley:
…uild

*Issue #, if available:*

*Description of changes:*


matthewfala (approved):

ShelbyZ:
For the shell files can you consider adding the executable flag via git:

```
git add --chmod=+x path/to/file
```
This will ensure that anyone pulling these changes down will not need to do `chmod +x` to set the executable flag later.

troubleshooting/tutorials/cloud-firelens-crash-repro-template/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM {core dump image build}
+ADD extra.conf /extra.conf
+ADD parser.conf /parser.conf
```
ShelbyZ, line 3: This is referenced only here and vaguely in the runbooks details. If we plan on providing a parser file we should consider adding it similar to extra.conf with a comment to fill in

PettitWesley: So I realized this and then decided to keep it on purpose, so that if you just run `docker build` without thinking it will fail. 

But yea I guess it still serves as a reminder and is less annoying if I just add a parser.conf in the template...

PettitWesley:
> For the shell files can you consider adding the executable flag via git:

@ShelbyZ none of the shell files in the new example are meant to be run outside of containers.

troubleshooting/debugging.md
```diff
@@ -674,6 +679,95 @@ To work around this issue, we suggest using the following option so that system
...
+
+In [troubleshooting/tutorials/cloud-firelens-crash-repro-template](tutorials/cloud-firelens-crash-repro-template/), you will find a template for setting up a customer repro in Fargate. 
+
+This can help you quickly setup a repro. The template and this guide require thought and consideration to create an ideal replication of the customer's setup.
+
+Perform the following steps:
+1. Copy the template to a new working directory.
+2. Setup the custom logger. Note: this step is optional, and other tools like [firelens-datajet](https://github.com/aws/firelens-datajet) may be used instead. You need some simulated log producer.
+    1. If the customer tails a file, add their log content or simulated log content to the `file1.log`. If they tail more than 1 file, then add a `file2.log` and add an entry in the `logger.sh` for it. If they do not tail a file, remove the file entry in `logger.sh`.
+    2. If the customer has a TCP input, place their log content or simulated log content in `tcp.log`. If there are multiple TCP inputs or no TCP inputs, customize `logger.sh` accordingly.
```
ShelbyZ, line 755: TCP could be more explicit that additional TCP listeners will need additional lines in the shell script to account for it and environment variables or hardcoded values in the task definition

ShelbyZ:
I wonder if we could do a Cloudformation YAML/JSON that included the task definition and referenced parameters ( we need to change provide so that we can automate some of the fill in the details for Task Definition. Values like environment variables and such can be replaced via overrides

troubleshooting/debugging.md
```diff
@@ -674,6 +679,95 @@ To work around this issue, we suggest using the following option so that system
...
 ```
 
+### Runbooks
+
+
+#### FireLens Crash Report Runbook
+
+When you recieve a SIGSEGV/crash report from a FireLens customer, perform the following steps. 
+
+##### 1. Build and distribute a core dump S3 uploader image
```
ShelbyZ, line 689: I wonder if we need details about using --platform and targeting the desired architecture when working on MacOS (building linux/amd64 images is not default). I am not sure the correct location to state that as make core will build whatever platform default and any guide information on building an image below may also need that note.

PR #548, opened 2023-02-17
Reorganize Dockerfiles and add init debug suite

matthewfala:
🌈

Previously we had alot of code duplication in dockerfiles making maintenance of all the debug and init image permutations difficult.  Updating docker files each time we want to build s3 or debug-init-s3 is not sustainable. This PR breaks up build logic into composable pieces and updates the makefile directives.

Builds on https://github.com/PettitWesley/aws-for-fluent-bit/tree/core-upload-2_29_0

Here's a dependency structure of the changes for a mental picture of what docker files depend on each other.
```
build (l1)
    ├── main-debug-base (l1)
    │   ├── 1.\ rebase-to-amazonlinux
    │   ├── init-debug-base (l2)
    │   │   ├── 1.\ dep-build-init
    │   │   ├── init-debug-fs (l3)
    │   │   └── init-debug-s3 (l3)
    │   ├── main-debug-fs (l2)
    │   ├── main-debug-s3 (l2)
    │   └── main-debug-valgrind (l2)
    └── main-release (l1)
        ├── 1.\ rebase-to-amazonlinux
        └── init-release (l2)
            └── 1.\ dep-build-init
```

One concern is that init-release goes from 1 layer image to a 2 layer image. Not sure if that's alright with us.


Makefile
```diff
@@ -13,12 +13,23 @@
...
 .PHONY: release
-release: linux-plugins
+release: build
 	docker system prune -f
-	docker build -t amazon/aws-for-fluent-bit:latest -f Dockerfile .
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:latest -f Dockerfile.vanilla-release .
+	docker system prune -f
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:init-latest -f Dockerfile.init-release .
+
+.PHONY: debug
```
PettitWesley, line 26: this issue noted that we need to update debugging guide when we have a finalized this, since you can't just build the dockerfiles anymore: https://github.com/aws/aws-for-fluent-bit/issues/579

scripts/dockerfiles/Dockerfile.main-debug-s3
```diff
@@ -0,0 +1,51 @@
...
+        echo "Note: Please set S3_KEY_PREFIX environment variable to a useful identifier - e.g. company name, team name, customer name"; \
+    fi; \
+    export RANDOM_ID_VALUE=$(($RANDOM%99999))$(($RANDOM%99999))$(($RANDOM%99999)); \
+    echo "RANDOM_ID is set to $RANDOM_ID_VALUE"; \
+    /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf; \
+    ls /cores-out/core*; \
+    export CORE_FILENAME=`basename $S3_KEY_PREFIX`_`date +"%FT%H%M%S"`${HOSTNAME+_host-$HOSTNAME}_${RANDOM_ID_VALUE}; \
+    mv `ls /cores-out/core*` ${CORE_FILENAME}.core; \
+    cp /fluent-bit/bin/fluent-bit ./${CORE_FILENAME}.executable; \
+    gdb -batch -ex 'thread apply all bt full' -ex 'quit' '/fluent-bit/bin/fluent-bit' ${CORE_FILENAME}.core > /cores-out/${CORE_FILENAME}.stacktrace \
```
PettitWesley, line 43: Is there any chance this could fail? Or any of the other commands? I'm wondering if we should consider the ordering of these based on what might fail to ensure we get something useful

scripts/dockerfiles/Dockerfile.main-debug-s3
```diff
@@ -0,0 +1,51 @@
...
+    zip -r /${CORE_FILENAME}.all.zip /cores-out; \
+    mv /${CORE_FILENAME}.all.zip /cores-out/${CORE_FILENAME}.all.zip; \
+    zip /cores-out/${CORE_FILENAME}.core.zip /cores-out/${CORE_FILENAME}.core; \
+    rm /cores-out/${CORE_FILENAME}.core; \
+    cp /cores-out/* /cores/; \
+    aws s3 cp /cores-out s3://${S3_BUCKET}/${S3_KEY_PREFIX}/ --recursive; \
+    echo "Stacktrace - ${CORE_FILENAME}.stacktrace:"; \
+    cat /cores/${CORE_FILENAME}.stacktrace
```
PettitWesley, line 51: Really cool that you figured out how to make this work 👏 👏

scripts/dockerfiles/Dockerfile.main-debug-s3
matthewfala: Hmm that's a good thought. Let me update this.

matthewfala: I updated the code accordingly and pulled it out into a separate file to reduce code duplication.

Makefile
```diff
@@ -13,12 +13,28 @@
...
+
+.PHONY: build
+build: linux-plugins
 	docker system prune -f
-	docker build -t amazon/aws-for-fluent-bit:init-latest -f Dockerfile.init .
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:build -f ./scripts/dockerfiles/Dockerfile.build .
+
+.PHONY: build-init
+release: build-init
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:build-init      -f ./scripts/dockerfiles/Dockerfile.build-init .
```
matthewfala, line 37: remove this space

scripts/dockerfiles/Dockerfile.init-debug-base
```diff
@@ -0,0 +1,14 @@
+
```
matthewfala, line 1: remove this space

scripts/dockerfiles/Dockerfile.init-release
```diff
@@ -0,0 +1,13 @@
...
+
+RUN mkdir -p /init
+
+COPY --from=init-builder /fluent_bit_init_process /init/fluent_bit_init_process
+
+COPY init/fluent_bit_init_entrypoint.sh /init/fluent_bit_init_entrypoint.sh
+RUN chmod +x /init/fluent_bit_init_entrypoint.sh
+
+# Only last CMD command will be executed, automatically replaces the original entrypoint
+CMD /init/fluent_bit_init_entrypoint.sh
```
matthewfala, line 13: add line space at end

Makefile
```diff
@@ -13,12 +13,28 @@
...
 
+# Improve build speeds during development by removing the --no-cache flag
+DOCKER_BUILD_FLAGS=--no-cache
+
 .PHONY: release
-release: linux-plugins
+release: build build-init
+	docker system prune -f
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:main-release -f ./scripts/dockerfiles/Dockerfile.main-release .
+	docker tag amazon/aws-for-fluent-bit:main-release amazon/aws-for-fluent-bit:latest
```
PettitWesley, line 23: why the extra tagging step? Why not just directly build it and tag it as `latest`??

Makefile
```diff
@@ -48,22 +65,26 @@ linux-plugins:
...
-	docker system prune -f
-	docker build -t amazon/aws-for-fluent-bit:valgrind -f Dockerfile.core .
+    	--CLOUDWATCH_PLUGIN_BRANCH=${CLOUDWATCH_PLUGIN_BRANCH} \
+		--DOCKER_BUILD_FLAGS=${DOCKER_BUILD_FLAGS}
+
+.PHONY: main-debug
+main-debug: main-debug-base
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:debug-fs       -f ./scripts/dockerfiles/Dockerfile.main-debug-fs .
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:debug-s3       -f ./scripts/dockerfiles/Dockerfile.main-debug-s3 .
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:debug-valgrind -f ./scripts/dockerfiles/Dockerfile.main-debug-valgrind .
```
PettitWesley, line 75: I think this is fine, but I was kind of expecting separate targets to build each type of debug image. Say you just want valgrind for example.

integ/integ.sh
```diff
@@ -2,7 +2,7 @@
 export AWS_REGION="us-west-2"
 export PROJECT_ROOT="$(pwd)"
 export VOLUME_MOUNT_CONTAINER="/out"
-export ValidateS3Dockerfile="Dockerfile"
+export ValidateS3Dockerfile="./scripts/dockerfiles/Dockerfile"
```
PettitWesley, line 5: I think we should keep the integ dockerfiles inside the integ folder. Also is this actually the path?

integ/run-integ.ps1
```diff
@@ -80,7 +80,7 @@ $env:AWS_FOR_FLUENT_BIT_CONTAINER_NAME = "aws-for-fluent-bit-$($env:TAG)"
 $env:ARCHITECTURE= "x86-64"
 $env:LOG_GROUP_NAME="fluent-bit-integ-test-$Platform-$env:ARCHITECTURE"
 $env:VOLUME_MOUNT_CONTAINER="C:/out"
-$env:ValidateS3Dockerfile = "Dockerfile.windows"
+$env:ValidateS3Dockerfile = "./scripts/dockerfiles/Dockerfile.windows"
```
PettitWesley, line 83: same comment as above

scripts/core_uploader.sh
```diff
@@ -0,0 +1,97 @@
...
+
+
+export S3_BUCKET=$1
+export S3_KEY_PREFIX=$2
+export CRASH_FOLDER="/cores-out"
+
+# Folder Structure
+# $CRASH_FOLDER: All the crash symbol information is combined and added to this folder.
+#    It is uploaded to s3 and added to the /cores folder
+# /cores: Files are sent to the cores folder
```
PettitWesley, line 29: but its now /cores-out ?

scripts/core_uploader.sh
```diff
@@ -0,0 +1,97 @@
...
+export CORE_FILENAME=`basename $S3_KEY_PREFIX`_`date +"%FT%H%M%S"`${HOSTNAME+_host-$HOSTNAME}_${RANDOM_ID_VALUE}
+
+# Check if there is a corefile
+export NO_CORE_STR="ls: cannot access $CRASH_FOLDER/core*: No such file or directory"
+export LS_CORE_STR="$(ls $CRASH_FOLDER/core* 2>&1 >/dev/null)"
+
+if [ "$NO_CORE_STR" == "$LS_CORE_STR" ]; then
+
+    # No corefile
+    echo "No core file to upload"
```
PettitWesley, line 43: I like this, so we can tell if it crashed and didn't produce a core

scripts/core_uploader.sh
```diff
@@ -0,0 +1,97 @@
...
+
+    # $CRASH_FOLDER
+    #   - CORE_FILENAME.core.zip
+    #   - CORE_FILENAME.core
+
+    # Upload core zip
+    aws s3 cp $CRASH_FOLDER/${CORE_FILENAME}.core.zip s3://${S3_BUCKET}/${S3_KEY_PREFIX}/
+
+    # Remove corefile zip to root folder
+    mv $CRASH_FOLDER/${CORE_FILENAME}.core.zip /tmp/${CORE_FILENAME}.core.zip
```
PettitWesley, line 57: why?

scripts/core_uploader.sh
```diff
@@ -0,0 +1,97 @@
...
+    cp $CRASH_FOLDER/* /cores/
+
+    # Send crash files to S3
+    aws s3 cp $CRASH_FOLDER s3://${S3_BUCKET}/${S3_KEY_PREFIX}/ --recursive
+
+    # Log stacktrace
+    echo "Stacktrace - ${CORE_FILENAME}.stacktrace:"
+    cat /cores/${CORE_FILENAME}.stacktrace
+
+    # Delay for 1 minute to allow for slow EFS uploads to complete if used
```
PettitWesley, line 94: explain EFS core file transfers?? once the AWS cp command is exited we've uploaded right?

scripts/dockerfiles/Dockerfile.build
```diff
@@ -28,6 +28,7 @@ RUN yum install -y  \
       pkgconfig \
       systemd-devel \
       zlib-devel \
+      valgrind-devel \
```
PettitWesley, line 31: our main build dockerfile should not need to include valgrind... am I misunderstanding this?

scripts/dockerfiles/Dockerfile.init-debug-s3
```diff
@@ -0,0 +1,47 @@
...
+        echo "Note: Please set S3_BUCKET environment variable to your crash symbol upload destination S3 bucket"; \
+    fi; \
+    if [ "$S3_KEY_PREFIX" == "issue" ]; then \
+        echo "Note: Please set S3_KEY_PREFIX environment variable to a useful identifier - e.g. company name, team name, customer name"; \
+    fi; \
+    /init/fluent_bit_init_process; \
+    chmod +x /init/invoke_fluent_bit.sh; \
+    export RANDOM_ID_VALUE=$(($RANDOM%99999))$(($RANDOM%99999))$(($RANDOM%99999)); \
+    echo "RANDOM_ID is set to $RANDOM_ID_VALUE"; \
+    env RANDOM_ID=$RANDOM_ID_VALUE /init/invoke_fluent_bit.sh; \
```
PettitWesley, line 46: why create and echo the random ID on startup? This will confuse user. We should only print this when we crash??

scripts/dockerfiles/Dockerfile.main-debug-base
```diff
@@ -0,0 +1,71 @@
+FROM amazon/aws-for-fluent-bit:build as builder
+COPY ./scripts/dockerfiles/Dockerfile.build /Dockerfile.1.build
```
PettitWesley, line 2: why are we copying one dockerfile inside another??

scripts/dockerfiles/Dockerfile.main-debug-fs
```diff
@@ -0,0 +1,4 @@
+FROM amazon/aws-for-fluent-bit:main-debug-base
+COPY ./scripts/dockerfiles/Dockerfile.main-debug-base /Dockerfile.2.main-debug-base
+
+CMD /fluent-bit/bin/fluent-bit -c /fluent-bit/etc/fluent-bit.conf; echo "Waiting 2 minutes for EFS core file transfers"; sleep 120
```
PettitWesley, line 4: Oh so the core files weren't working on EFS because they are big and take a while???

Makefile
matthewfala: That's probably a good idea. I can change it.

scripts/core_uploader.sh
matthewfala: Right... So if you mount /cores to some EFS Volume, the problem is that there may be other cores already there, so you can't expect that folder to be empty

scripts/core_uploader.sh
matthewfala: We don't want to zip the core zip into the core file.

scripts/core_uploader.sh
matthewfala: Right... So in my testing if the customer mounts an EFS volume to cores and you exit immediately, the core file doesn't actually get transfered to EFS... I needed to add a delay and then it seems like the volume gets synchronized

scripts/dockerfiles/Dockerfile.build
matthewfala: Does this picture help:
```
build
    ├── main-debug-base
    │   ├── 1.\ rebase-to-amazonlinux
    │   ├── init-debug-base
    │   │   ├── 1.\ dep-build-init
    │   │   ├── init-debug-fs
    │   │   └── init-debug-s3
    │   ├── main-debug-fs
    │   ├── main-debug-s3
    │   └── main-debug-valgrind
    └── main-release
        ├── 1.\ rebase-to-amazonlinux
        └── init-release
            └── 1.\ dep-build-init
```

Makefile
matthewfala: They build super quickly, so I though why not just put all of them together. The part that takes a long time to build is the main-debug-base

scripts/dockerfiles/Dockerfile.main-debug-fs
matthewfala: That's right! In my testing the core files wouldn't actually get synced to EFS unless a delay was added.

scripts/dockerfiles/Dockerfile.main-debug-base
matthewfala: Just for documentation on what else the image is based on... I can get rid of this if you want

scripts/dockerfiles/Dockerfile.init-debug-s3
matthewfala: It's because you can use the random id in your Fluent Bit configuration ie send logs to a cloudwatch log  stream with the random id. Then you have a way of connecting a fluent bit instance with a specific log stream

scripts/dockerfiles/Dockerfile.build
matthewfala: Actually, if DFLB_VALGRIND compiler flag is set you may need valgrind-devel as a dep.

https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile.debug-base#L31

So you do need valgrind-devel

The deps in the build image are not actually included in any of the final images

Makefile
matthewfala: Actually, it's because the Dockerfile.init-release depends on main-release. That's why we build main-release, and tag with latest

integ/run-integ.ps1
matthewfala: My mistake. I didn't understand the context of this. Reverting

Makefile
PettitWesley: yea its fine

Makefile
```diff
@@ -13,12 +13,28 @@
 
 all: release
 
+# Improve build speeds during development by removing the --no-cache flag
+DOCKER_BUILD_FLAGS=--no-cache
```
PettitWesley, line 17: so this is just a nicety so if I want to remove `--no-cache` I edit the makefile in one place instead of multiple + script?

scripts/core_uploader.sh
PettitWesley: Sorry I don't understand what you are saying. Does this change the EFS core tutorial I wrote or not?

scripts/core_uploader.sh
PettitWesley: ah

scripts/core_uploader.sh
PettitWesley: interesintg. I guess I never actually tested that now that I think about... I only did the live debugging for running frozen task core file.

scripts/dockerfiles/Dockerfile.build
PettitWesley: But wait. For non-debug build, should we enable -DFLB_VALGRIND? I think no. 

For the debug images, I think we need valgrind-devel at runtime for valgrind to work...

scripts/core_uploader.sh
matthewfala: No, https://github.com/aws/aws-for-fluent-bit/pull/548/files#diff-ac4a590fd0373a499e5f24246e9c6b554b57cf458124af8e48d471ed9ba48128R85. We do all the processing in the /cores-out folder, but in the end we copy all the files to the /cores folder which is what you might mount an EFS volume to...


The goal is to make sure that this folder is empty when you start. If you mount an efs volume, it might not be empty.

Makefile
matthewfala: That's right. you can change all the docker build flags in one place

scripts/dockerfiles/Dockerfile.build
matthewfala: https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile.debug-base#L31

Currently it is featured in the debug images:
https://github.com/aws/aws-for-fluent-bit/blob/mainline/Dockerfile.debug-base#L31

scripts/core_uploader.sh
matthewfala: Ah.... Yeah, a bit of an unexpected problem.

scripts/build_windows_fluent_bit.ps1
```diff
@@ -290,4 +290,4 @@ Copy-Item -Path "${AWSForFluentBitRootDir}\AWS_FOR_FLUENT_BIT_VERSION" -Destinat
 Copy-Item -Path "${AWSForFluentBitRootDir}\scripts\entrypoint.ps1" -Destination "${AWSForFluentBitRootDir}\build\windows\entrypoint.ps1"
 
 # Copy the dockerfile used to build an image. This would ensure that the images remains constant in time.
-Copy-Item -Path "${AWSForFluentBitRootDir}\Dockerfile.windows" -Destination "${AWSForFluentBitRootDir}\build\windows\Dockerfile.windows"
+Copy-Item -Path "${AWSForFluentBitRootDir}\scripts\dockerfiles\Dockerfile.windows" -Destination "${AWSForFluentBitRootDir}\build\windows\Dockerfile.windows"
```
matthewfala, line 293: Does this also need to change?

scripts/core_uploader.sh
PettitWesley: Ah... makes sense

scripts/dockerfiles/Dockerfile.init-debug-s3
PettitWesley: hmmm ok... if this is going to be a new feature.. we should document it somewhere... it is just for debug images or for prod release too?

matthewfala: It's just for the debug images, not in prod.

integ/integ.sh
matthewfala: Also, this is a great catch, thank you!

PettitWesley (approved):

PR #559, opened 2023-02-26
Add support for publishing a non-latest version

PettitWesley:
*Issue #, if available:*

*Description of changes:*


PettitWesley:
I've deployed the pipeline in our public-ecr test account. I'm running through multiplereleases to make it all works. 

Here's the result of the first one, where I set `latest: false`. As expected, it publishes images but without a latest tag.

I will now do a release with `latest: true`, then again one with false. If all works as expected then we are good to go. 

I'm also checking the sync tasks, so far for the existing case I can show that there is no regression. Haven't tested syncing a non-latest release yet. 

Screen Shot 2023-02-26 at 10 01 53 PM

ShelbyZ (approved):

scripts/get_latest_dockerhub_version.py
```diff
@@ -22,4 +26,12 @@
...
-print(numeric_tags[0])
+
+if check_version == "latest":
+    print(numeric_tags[0]) # print latest version
+else:
+    # prints true if the version is in the list but not latest
+    # printing true means publish.sh sync a non-latest version
+    is_latest = (numeric_tags[0] == check_version)
+    is_published = check_version in numeric_tags
+    print(str(not is_latest and is_published).lower()) # print if version is latest
```
matthewfala, line 37: Maybe remove this comment or specify:
- print true if version is not latest and already published

scripts/publish.sh
```diff
@@ -21,13 +21,33 @@ IMAGE_SHA_MATCHED="FALSE"
...
+
+# Problem: when we push a new version bump the version number in AWS_FOR_FLUENT_BIT_VERSION file changes
+# but that version is not published immediately. Thus, sync tasks normally
+# sync latest version found in DockerHub. But what if we want to release a non-latest version?
+# then sync tasks need to know this. So we use the script to check for an already published version
+# that's not latest
 docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
 tag_array=(`echo ${docker_hub_image_tags}`)
-AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py linux ${tag_array[@]})
+AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py linux latest ${tag_array[@]})
```
matthewfala, line 34: tag_array[@] only has the last 100 tags.

see
```
curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=1050' | jq -r '.results[].name' | wc -l
```

I think this should be okay for the given logic. I think more recent tags appear at the top, and so this means that if we  publish multiple times we catch the previous publish

scripts/get_latest_dockerhub_version.py
```diff
@@ -22,4 +26,12 @@
         numeric_tags.append(tag_to_store)
 
 numeric_tags.sort(key=lambda s: list(map(int, s.split('.'))), reverse=True)
-print(numeric_tags[0])
+
+if check_version == "latest":
+    print(numeric_tags[0]) # print latest version
+else:
```
matthewfala, line 32: A little confusing at first how this function now returns either tag name of latest version or true false, based on whether the "latest" check_version argument is used. But I like it. It makes sense after thinking about it.

scripts/publish.sh
```diff
@@ -466,15 +509,18 @@ verify_ecr() {
 		docker pull ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:"$init"-${AWS_FOR_FLUENT_BIT_VERSION}
 		sha1_init=$(docker inspect --format='{{index .RepoDigests 0}}' ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:"$init"-${AWS_FOR_FLUENT_BIT_VERSION})
 	fi
-	docker pull ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:latest
-	sha2=$(docker inspect --format='{{index .RepoDigests 0}}' ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:latest)
 
-	verify_sha $sha1 $sha2
+	if [ "${PUBLISH_LATEST}" = "true" ]; then
```
matthewfala, line 513: The non publish latest version doesn't get verified right? And that's okay?

scripts/publish.sh
```diff
@@ -251,7 +286,7 @@ check_parameter() {
 	fi
 }
 
-sync_latest_image() {
+sync_image_version() {
```
matthewfala, line 289: Will there ever be a case where the AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB is an empty string "" because we:

1. Merged the code
2. Have not finished the release pipeline yet
3. Sync tasks begin

Would there be some error in this case where the sync tasks fail?

matthewfala (commented):
Added some comments. Slightly confused about how often the sync task function is invoked and if any logic is broken if it is invoked when AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB is "" (which previously couldn't happen)

scripts/publish.sh
```diff
@@ -21,13 +21,33 @@ IMAGE_SHA_MATCHED="FALSE"
...
+# If the AWS_FOR_FLUENT_BIT_VERSION is an older version which is already published to dockerhub
+# and latest is set to false in linux.version, then we sync an older non-latest version.
+# otherwise, normal behavior, sync latest version found in dockerhub
+if [ "${PUBLISH_LATEST}" = "false" ]; then
+	PUBLISH_NON_LATEST=$(./get_latest_dockerhub_version.py linux ${AWS_FOR_FLUENT_BIT_VERSION} ${tag_array[@]})
+	if [ "${PUBLISH_NON_LATEST}" = "true" ]; then
+		AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=${AWS_FOR_FLUENT_BIT_VERSION}
+	fi
+fi
+
```
PettitWesley, line 45: @matthewfala I could add a check here to fail fast if the `AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB` is empty string. How could that case happen now though? I think it still shouldn't happen,

scripts/publish.sh
PettitWesley: Its still first set to the top version from the python script sorted list. And then it can be set to the version in the AWS_FOR_FLUENT_BIT version file instead if we are not publishing a latest release. So I think it can't be empty string. Do you see a case?

scripts/publish.sh
PettitWesley: ah good point... the logic should be improved here I think:

1. We should always verify the image version in the region against the one in public ECR.
2. If we are publishing a latest release, we should also check that the latest and version number images are the same.

PettitWesley: Number one is only for the sync task case. When in pipeline public ECR won't be published yet.

scripts/get_latest_dockerhub_version.py
PettitWesley: sure that's more clear

PettitWesley: I've also now added more description in code comment at the beginning of the script

scripts/publish.sh
PettitWesley: good point we are relying on behavior in DH API. I will add a comment on this

PettitWesley: We also only have 64 tags so far..

PettitWesley: we probably won't reach 100 tags till next year

PettitWesley: ECS Agent has more than 100 tags, and it does indeed print the latest tags:

```
147dda285fd4:aws-for-fluent-bit wppttt$ curl -s -S  'https://registry.hub.docker.com/v2/repositories/amazon/amazon-ecs-agent/tags/?page=1&page_size=250' | jq -r '.results[].name' | grep 1.69
v1.69.0
arm64-v1.69.0
amd64-v1.69.0
147dda285fd4:aws-for-fluent-bit wppttt$ curl -s -S  'https://registry.hub.docker.com/v2/repositories/amazon/amazon-ecs-agent/tags/?page=1&page_size=250' | jq -r '.results[].name' | grep 1.68
v1.68.2
arm64-v1.68.2
amd64-v1.68.2
v1.68.1
arm64-v1.68.1
amd64-v1.68.1
v1.68.0
arm64-v1.68.0
amd64-v1.68.0
147dda285fd4:aws-for-fluent-bit wppttt$ curl -s -S  'https://registry.hub.docker.com/v2/repositories/amazon/amazon-ecs-agent/tags/?page=1&page_size=250' | jq -r '.results[].name' | grep 1.1
147dda285fd4:aws-for-fluent-bit wppttt$ curl -s -S  'https://registry.hub.docker.com/v2/repositories/amazon/amazon-ecs-agent/tags/?page=1&page_size=250' | jq -r '.results[].name' | grep 1.15
```

scripts/publish.sh
PettitWesley: I have updated the script to always check the version number tag in local region ECR against public ECR. If we are publishing latest, we also check against in region latest.

scripts/publish.sh
matthewfala: I actually don't see any way that AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB can be an empty string any more. It will be latest if not the AWS_FOR_FLUENT_BIT_VERSION as you describe in your comment below.

scripts/publish.sh
matthewfala: I actually don't see any way that AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB can be an empty string any more. It will be latest if not the AWS_FOR_FLUENT_BIT_VERSION as you describe in your comment

scripts/publish.sh
matthewfala: This looks good!

PettitWesley: actually we can't verify against public ECR outside of sync task...

scripts/publish.sh
```diff
@@ -471,22 +517,48 @@ verify_ecr() {
...
+		verify_sha $sha1_init $sha2_init
 	else
 		docker pull ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_VERSION}
 		sha1=$(docker inspect --format='{{index .RepoDigests 0}}' ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_VERSION})
 
 		docker pull ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:"$init"-${AWS_FOR_FLUENT_BIT_VERSION}
 		sha1_init=$(docker inspect --format='{{index .RepoDigests 0}}' ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:"$init"-${AWS_FOR_FLUENT_BIT_VERSION})
+
+		# verify version number tag against public ECR
+		docker pull public.ecr.aws/aws-observability/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_VERSION}
```
PettitWesley, line 539: this doesn't work... you can't verify against public ECR outside of sync task in main pipeline before you've published to public ECR...

scripts/publish.sh
```diff
@@ -21,13 +21,36 @@ IMAGE_SHA_MATCHED="FALSE"
...
+# this code currenly works because DockerHub returns the only last 100 tags and as of March 2023 we only have 64
+# and it should keep working because dockerhub returns the latest tags first
 docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
 tag_array=(`echo ${docker_hub_image_tags}`)
-AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py linux ${tag_array[@]})
+AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py linux latest ${tag_array[@]})
+
+# If the AWS_FOR_FLUENT_BIT_VERSION is an older version which is already published to dockerhub
+# and latest is set to false in linux.version, then we sync an older non-latest version.
+# otherwise, normal behavior, sync latest version found in dockerhub
```
PettitWesley, line 41: I can't remember now why I didn't just do the work to make it so that the sync tasks sync any latest images that are in docker hub but not in the repo.

PR #572, opened 2023-03-03
Add manifest if dockerhub not equal to stable

ShelbyZ:
*Description of changes:*
- fix spacing on line240
- check stable vs dockerhub to see if manifest "might" be missing

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -237,7 +237,7 @@ check_parameter() {
 	repo_uri=$(sed -e 's/^"//' -e 's/"$//' <<<"$repo_uri")
 	docker pull $repo_uri
 
-	if [ "${2}" != "stable"]; then 
+	if [ "${2}" != "stable" ]; then 
 		repo_uri_init=$(aws ssm get-parameter --name /aws/service/aws-for-fluent-bit/"$init"-${2} --region ${1} --query 'Parameter.Value')
 		IFS='.' read -r -a array <<<"$repo_uri_init"
 		region="${array[3]}"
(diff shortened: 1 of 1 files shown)


Claych (approved):

scripts/publish.sh
```diff
@@ -309,6 +309,11 @@ sync_latest_image() {
 		create_manifest_list_init ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit "init-latest" ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
 		create_manifest_list_init ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit "$init"-${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB} ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
 	fi
+
+	if [ "${AWS_FOR_FLUENT_BIT_STABLE_VERSION}" != "${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}" ]; then
+		create_manifest_list ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_STABLE_VERSION} ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
```
PettitWesley, line 314: just checking, creating the manifest for the stable version tag over and over again is safe right? I think it is..

ShelbyZ: If you were to create the manifest locally for the same tag twice you would be alerted to use a the `--amend` flag. Since this is handled via a serverless solution we should not have leftover manifest locally. After the local manifest is created the push command will update what is on the repo and I have yet to get that to complain about a duplicate.

PR #573, opened 2023-03-03
Allow failover of SSM query when namespace has never been used

ShelbyZ:
*Description of changes:*
- fix naming of imageTag --> tagCount for readability
- Provide helper method, sync_ssm, to handle checking/fallback logic with SSM Parameter Store with the following behavior

> If value is not present, we would receive error (set -e) and exit
> Instead we use `!` to check if the command failed and set a prepare boolean
> If the command succeeds we follow the normal logic to check if the request had invalid parameters
> If we should prepare the namespace or had invalid parameters we should attempt to put a variable in the namespace
> If we are an owner the request will succeed and the namespace is prepared
> If we are not an owner the request will fail with access denied and exit immediately

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -198,25 +198,26 @@ publish_to_public_ecr() {
 }
 
 publish_ssm() {
-	aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/${3} --overwrite \
-		--description 'Regional Amazon ECR Image URI for the latest AWS for Fluent Bit Docker Image' \
-		--type String --region ${1} --value ${2}:${3}
-	aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/latest --overwrite \
-		--description 'Regional Amazon ECR Image URI for the latest AWS for Fluent Bit Docker Image' \
-		--type String --region ${1} --value ${2}:latest
-	
-	aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/"$init"-${3} --overwrite \
-		--description 'Regional Amazon ECR Image URI for the "$init"-latest AWS for Fluent Bit Docker Image' \
-		--type String --region ${1} --value ${2}:"$init"-${3}
-	aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/"$init"-latest --overwrite \
-		--description 'Regional Amazon ECR Image URI for the "$init"-latest AWS for Fluent Bit Docker Image' \
-		--type String --region ${1} --value ${2}:"$init"-latest
-}
-
-publish_stable_ssm() {
-	aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/stable --overwrite \
-		--description 'Regional Amazon ECR Image URI for the latest stable AWS for Fluent Bit Docker Image' \
-		--type String --region ${1} --value ${2}:${3}
+	# This optional parameter indicates if we should publish stable (defaults to false)
+	if [ ${4:-false} = true ]; then
+		aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/stable --overwrite \
+			--description 'Regional Amazon ECR Image URI for the latest stable AWS for Fluent Bit Docker Image' \
+			--type String --region ${1} --value ${2}:${3}
+	else
+		aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/${3} --overwrite \
+			--description 'Regional Amazon ECR Image URI for the latest AWS for Fluent Bit Docker Image' \
+			--type String --region ${1} --value ${2}:${3}
+		aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/latest --overwrite \
+			--description 'Regional Amazon ECR Image URI for the latest AWS for Fluent Bit Docker Image' \
+			--type String --region ${1} --value ${2}:latest
+		
+		aws ssm put-parameter --name /aws/service/aws-for-fluent-bit/"$init"-${3} --overwrite \
+			--description 'Regional Amazon ECR Image URI for the "$init"-latest AWS for Fluent Bit Docker Image' \
+			--type String --region ${1} --value ${2}:"$init"-${3}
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -318,13 +318,13 @@ sync_latest_image() {
...
 
-	ssm_parameters=$(aws ssm get-parameters --names "/aws/service/aws-for-fluent-bit/${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}" --region ${region})
+	ssm_parameters=$(aws ssm get-parameters --names "/aws/service/aws-for-fluent-bit/${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}" --region ${region} || true)
 	invalid_parameter=$(echo $ssm_parameters | jq .InvalidParameters[0])
 	if [ "$invalid_parameter" != 'null' ]; then
 		publish_ssm ${region} ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
 	fi
 
-	ssm_parameters=$(aws ssm get-parameters --names "/aws/service/aws-for-fluent-bit/stable" --region ${region})
+	ssm_parameters=$(aws ssm get-parameters --names "/aws/service/aws-for-fluent-bit/stable" --region ${region} || true)
```
PettitWesley, line 327: Actually I don't see how this is better, this means the script will continue on if we are not allow-listed for SSM rather than just failing fast? 

Isn't that not ideal? Isn't failing fast better? Or am I misunderstanding?

PettitWesley (changes requested):

scripts/publish.sh
PettitWesley: The existing code handles missing parameters well via invalid parameters return:

```
147dda285fd4:aws-for-fluent-bit wppttt$ aws ssm get-parameters --names "/tmp/service/aws-for-fluent-bit/stable" --region us-west-2
{
    "Parameters": [],
    "InvalidParameters": [
        "/tmp/service/aws-for-fluent-bit/stable"
    ]
}
147dda285fd4:aws-for-fluent-bit wppttt$ echo $?
0
```

It does not handle invalid public parameters well, but we want that, to tell us if we messed up some setup:

```
147dda285fd4:aws-for-fluent-bit wppttt$ aws ssm get-parameters --names "/aws/service/does-not-exit/stable" --region us-west-2

An error occurred (AccessDeniedException) when calling the GetParameters operation: No access to "/aws/" namespace: aws/service/does-not-exit is not a valid namespace
147dda285fd4:aws-for-fluent-bit wppttt$ echo $?
254
```

So I think the existing code seems better?

PettitWesley (changes requested):

PettitWesley (approved):
Thank you for providing me with the answer that the parameter namespace does not exist until we put our first parameter meaning that my logic in the comments was not correct. Now I am fine with this change.

PettitWesley (changes requested):
As discussed, this current state is not what we want, we want to do one of the following:

1. Keep script as-is, if not allow listed or parameter namespace not prepared, then script blows up and the runbook tells you what to do to fix it.
2. Script handles the error exit code and then tries to put the parameter which then either succeeds or fails. Runbook handles failure case.

PettitWesley (changes requested):

scripts/publish.sh
```diff
@@ -283,6 +283,26 @@ sync_public_and_repo() {
...
+	should_prepare=false
+	# Check to see if namespace is prepared, once a parameter is put into namespace requests should not fail
+	if ! ssm_parameters=$(aws ssm get-parameters --names ${namespace_path} --region ${region}); then
+		should_prepare=true
+	else
+		invalid_parameter=$(echo $ssm_parameters | jq .InvalidParameters[0])
+	fi
+
+	if [ $should_prepare = true ] || [ "$invalid_parameter" != 'null' ]; then
+		publish_stable_ssm ${region} ${repo} ${tag}
```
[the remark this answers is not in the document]
ShelbyZ: Good catch, I'll rework that to handle the publish correctly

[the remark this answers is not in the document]
PettitWesley: neither of the existing functions seem ideal for this purpose from a quick glance

scripts/publish.sh
```diff
@@ -318,23 +345,14 @@ sync_latest_image() {
...
-	fi
+	sync_ssm "/aws/service/aws-for-fluent-bit/${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}" ${region} ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB}
+	sync_ssm "/aws/service/aws-for-fluent-bit/stable" ${region} ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
 
 	stable_uri=$(aws ssm get-parameters --names /aws/service/aws-for-fluent-bit/stable --region ${region} --query 'Parameters[0].Value')
 	stable_uri=$(sed -e 's/^"//' -e 's/"$//' <<<"$stable_uri")
 
 	if [ "$stable_uri" != "${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit:${AWS_FOR_FLUENT_BIT_STABLE_VERSION}" ]; then
-		publish_stable_ssm ${region} ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
+		publish_stable_ssm ${region} ${account_id}.dkr.ecr.${region}.${endpoint}/aws-for-fluent-bit ${AWS_FOR_FLUENT_BIT_STABLE_VERSION} true
```
PettitWesley, line 355: this should be publish_ssm now, since you removed publish_stable_ssm?

PettitWesley (changes requested):

PettitWesley (approved):

PR #577, opened 2023-03-07
debugging: add note that both CW plugins consume retries on create gr…

PettitWesley:
…oup/stream failures

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -44,6 +44,7 @@
     - [kinesis_streams or kinesis_firehose partially succeeded batches or duplicate records](#kinesisstreams-or-kinesisfirehose-partially-succeeded-batches-or-duplicate-records)
     - [Always use multiline in the tail input](#always-use-multiline-in-the-tail-input)
     - [Tail Input duplicates logs during rotation](#tail-input-duplicates-logs-because-of-rotation)
+    - [Both CloudWatch Plugins: Create failures consume retries](#both-cloudwatch-plugins-create-failures-consume-retries)
 - [Best Practices]
     - [Set Aliases on all Inputs and Outputs](#set-aliases-on-all-inputs-and-outputs)
     - [Tail Config with Best Practices](#tail-config-with-best-practices)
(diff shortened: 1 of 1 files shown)


Claych (approved):

troubleshooting/debugging.md
```diff
@@ -630,6 +631,18 @@ Path                /logs/request.log
 
 In other cases, the original config would be correct. For example, assume your app produces first a log file `request.log` and then next creates and writes new logs to a new file named `request.log1` and then next `request.log2`. And then when the max number of files are reached, the original `request.log` is deleted and then a new `request.log` is written. If that is the scenario, then the original Path pattern with a wildcard at the end is correct since it is necessary to match all log files where new logs are written. This method of log rotation is the behavior of the [log4j DirectWriteRolloverStrategy](https://logging.apache.org/log4j/2.x/manual/appenders.html#DirectWriteRolloverStrategy) where there is no file renaming. 
 
+#### Both CloudWatch Plugins: Create failures consume retries
+
+In Fluent Bit, [retries are handled by the core engine](https://docs.fluentbit.io/manual/administration/scheduling-and-retries). Whenever the plugin encounters any error, it returns a retry to the engine which schedules a retry. This means that log group creation, log stream creation or log retention policy calls can consume a retry if they fail.
```
ShelbyZ, line 636: Do these calls only happen on startup? Would they only effect the first batch of logs sent?

troubleshooting/debugging.md
```diff
@@ -630,6 +631,18 @@ Path                /logs/request.log
...
+
+*This applies to both the [CloudWatch Go Plugin and the CloudWatch C Plugin](#aws-go-plugins-vs-aws-core-c-plugins).*
+
+This means that before the plugin even tries to send a chunk of log records, it may consume a retry on a random temporary networking error trying to call CreateLogStream.
```
ShelbyZ, line 640: I don't know if it would feel clearer to provide an example scenario with default or set number of retries.

ShelbyZ (approved):

troubleshooting/debugging.md
PettitWesley: uhhh good point I will add a note. Depends on your config. With templating you can get lots of new groups and streams.

troubleshooting/debugging.md
PettitWesley: I think that just makes it too long

PR #585, opened 2023-03-11
core dump tutorial: add note on how to use and install debuginfo-install

PettitWesley:
*Issue #, if available:*

*Description of changes:*


ShelbyZ (approved):
`yum` it's correct for Amazon Linux and RHEL variants which I would guess is most our users. Debian and others might need to use different options

Claych (approved):

PR #602, opened 2023-03-22
add make local to build the image locally

Claych:
*Issue #, if available:*

*Description of changes:*


Dockerfile.init
```diff
@@ -15,7 +15,10 @@ ENV PATH ${PATH}:/home/.gimme/versions/gostable.linux.arm64/bin:/home/.gimme/ver
 RUN go version
 
 ENV GO111MODULE on
-RUN go env -w GOPROXY=direct
+
+ARG BUILD_TYPE="RELEASE"
+
+RUN if [ "$BUILD_TYPE"="LOCAL"  ];then go env -w GOPROXY=direct;fi
```
PettitWesley, line 21: typically I see space between the = and the string in bash: https://stackoverflow.com/questions/2237080/how-to-compare-strings-in-bash

I am not sure if this is required or not.

Claych: yes, this is really good point! But the problem so far is we can't pass BUILD_TYPE=LOCAL to here

Dockerfile.plugins
```diff
@@ -27,6 +26,9 @@ ARG FIREHOSE_PLUGIN_BRANCH=""
 ARG CLOUDWATCH_PLUGIN_CLONE_URL=https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit.git
 ARG CLOUDWATCH_PLUGIN_TAG=""
 ARG CLOUDWATCH_PLUGIN_BRANCH=""
+ARG BUILD_TYPE="RELEASE"
+
+RUN if [ "$BUILD_TYPE"="LOCAL"  ];then go env -w GOPROXY=direct;fi
```
PettitWesley, line 31: The docker build for this is called here: https://github.com/aws/aws-for-fluent-bit/blob/mainline/scripts/build_plugins.sh#L177

So you need to pass the var to there or else it won't work. 

Thank you @ShelbyZ for figuring this out.

PettitWesley: See here as well for setting different values for a Makefile var in different targets: https://github.com/aws/aws-for-fluent-bit/blob/mainline/Makefile#L24

Makefile
```diff
@@ -20,6 +20,13 @@ release: linux-plugins
 	docker system prune -f
 	docker build -t amazon/aws-for-fluent-bit:init-latest -f Dockerfile.init .
 
+.PHONY: local
+local: linux-plugins
+	docker system prune -f
+	docker build --build-arg LOCAL_BUILD=true -t amazon/aws-for-fluent-bit:latest -f Dockerfile .
```
PettitWesley, line 26: Do we need to pass  LOCAL_BUILD here? Its not used in the main `Dockerfile` I think

scripts/build_plugins.sh
```diff
@@ -174,6 +174,6 @@ fi
 
 if [ "$OS_TYPE" == "linux" ];
 then
-  docker build $PLUGIN_BUILD_ARGS --no-cache -t aws-fluent-bit-plugins:latest -f ./Dockerfile.plugins .
+  docker build --build-arg LOCAL_BUILD=true $PLUGIN_BUILD_ARGS --no-cache -t aws-fluent-bit-plugins:latest -f ./Dockerfile.plugins .
```
PettitWesley, line 177: How will this work in the cloud? Have you tested it there? I think this will always set the build flag no matter what. We discussed yesterday with @ShelbyZ how to pass an env var from the Makefile to this script to set this flag.

PettitWesley (changes requested):

Makefile
PettitWesley: Still have same question

scripts/build_plugins.sh
PettitWesley: Still have same question

PettitWesley: Sorry I missed the update with BUILD_TYPE

scripts/build_plugins.sh
```diff
@@ -172,8 +172,12 @@ then
   echo "Copied plugin archive to the build output folder"
 fi
 
-if [ "$OS_TYPE" == "linux" ];
+if [ "$OS_TYPE" == "linux" ] && [ "$BUILD_TYPE" == "local" ];
```
PettitWesley, line 175: do we need to use BUILD_TYPE for windows too??

PettitWesley (commented):
I think I can approve once you add windows

PR #603, opened 2023-03-26
add make local to build image locally

Claych:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,23 @@
 
 all: release
 
+local: export BUILD_TYPE = local
+release: export BUILD_TYPE = release
+
 # Improve build speeds during development by removing the --no-cache flag
 DOCKER_BUILD_FLAGS=--no-cache
 
 .PHONY: release
 release: build build-init
+	docker system prune -f
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:main-release -f ./scripts/dockerfiles/Dockerfile.main-release .
+	docker tag amazon/aws-for-fluent-bit:main-release amazon/aws-for-fluent-bit:latest
+	docker system prune -f
+	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:init-latest -f ./scripts/dockerfiles/Dockerfile.init-release .
+
+.PHONY: local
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -24,6 +27,15 @@ release: build build-init
 	docker system prune -f
 	docker build $(DOCKER_BUILD_FLAGS) -t amazon/aws-for-fluent-bit:init-latest -f ./scripts/dockerfiles/Dockerfile.init-release .
 
+.PHONY: local
+local: build
```
matthewfala, line 31: As Wesley commented earlier in discussion you may be able to use these environment variables to determine if you are in Code Build.

Maybe something like: `CODEBUILD_BUILD_ARN`. If this exists you know you're in the pipeline (presumably).

https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html

PettitWesley: this means it can be a single statement: RUN if -z blah then set env

Makefile
```diff
@@ -13,6 +13,9 @@
 
 all: release
 
+local: export BUILD_TYPE = local
+release: export BUILD_TYPE = release
+
```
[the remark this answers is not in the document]
PettitWesley: "usually" is not a good enough reason. Some day we will need to build windows locally, I am sure.

Makefile
matthewfala: Maybe use somehting like: if [-z $CODEBUILD_BUILD_ARN]; then export DOCKER_BUILD_FLAGS="LOCAL=true; $DOCKER_BUILD_FLAGS"; else export DOCKER_BUILD_FLAGS="LOCAL=false; $DOCKER_BUILD_FLAGS" fi

PettitWesley: Ah yea... I think you can't read env vars inside of DOckerfiles? I just tried testing that out myself and it didn't work. 

Otherwise I meant that inside the Dockerfile you could do:
```
RUN if [ -z "$CODEBUILD_BUILD_ARN" ];then go env -w GOPROXY=direct;fi

```

PettitWesley: Mine doesn't seem to work, so use @matthewfala suggestion.

PR #610, opened 2023-03-28
README: add public ECR auth error

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,13 @@ For example, you can pull the image with latest version by:
 docker pull public.ecr.aws/aws-observability/aws-for-fluent-bit:latest
 ```
 
-If you see errors for image pull limits, try log into public ECR with your AWS credentials:
+If you see errors for image pull limits, or get the following error:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -181,7 +181,13 @@ For example, you can pull the image with latest version by:
 docker pull public.ecr.aws/aws-observability/aws-for-fluent-bit:latest
 ```
 
-If you see errors for image pull limits, try log into public ECR with your AWS credentials:
+If you see errors for image pull limits, or get the following error:
```
ShelbyZ, line 184: Do we care to suggest running `docker logout public.ecr.aws` prior to this command or just re-login (as we know it has worked)?

ShelbyZ (approved):

matthewfala (approved):
Helpful. Thanks!

README.md
PettitWesley: I think login is better if you have an AWS account because IIRC there are pull limits outside of AWS datacenters... and login fixes that..

PR #619, opened 2023-03-31
init: fix S3 ARN parsing #617

### LINKED ISSUE #617: Init doesn't work in GovCloud
### Describe the question/issue

When using aws-for-fluentbit in the AWS GovCloud environment, loading from S3 using the ```init``` tagged images is not possible because of this line: https://github.com/aws/aws-for-fluent-bit/blob/mainline/init/fluent_bit_init_process.go#L205

Which assumes that the ARN prefix is always ```arn:aws:*``` (13 chars to the bucket name start), whereas, in GovCloud environments, this is actually ```arn:aws-us-gov:*``` (20 chars to the bucket name start)

This results in truncation on the wrong part, resulting in a target bucket/path of: 

### Configuration

Using envvars in ECS Fargate to load from S3; eg:
```
                {
                    "name": "aws_fluent_bit_init_s3_1",
                    "value": "arn:aws-us-gov:s3:::my_bucket_name/my_config_file.conf"
                },
```

### Fluent Bit Log Output 

Note the Bucket name starting with ```v:s3:::``` due to misaligned truncation
```
time="2023-03-30T16:16:17Z" level=fatal msg="[FluentBit Init Process] Cannot get bucket region of v:s3:::my_bucket_name + my_config_file.conf, you must be the bucket owner to implement this operation\n"
```

### Fluent Bit Version Info

2.31.5

###  Cluster Details

ECS Fargate

### Application Details

N/A

### Steps to reproduce issue

See above

### Related Issues

N/A

PettitWesley:
*Issue #, if available:*

Fixes #617 

*Description of changes:*

### DIFF
--- a/init/fluent_bit_init_process.go
+++ b/init/fluent_bit_init_process.go
@@ -195,17 +195,20 @@ func processConfigFile(path string) {
 	}
 }
 
-func getS3ConfigFile(arn string) string {
+func getS3ConfigFile(userInput string) string {
 	// Preparation for downloading S3 config files
 	if !s3ClientCreated {
 		createS3Client()
 	}
 
 	// e.g. "arn:aws:s3:::user-bucket/s3_parser.conf"
-	arnBucketFile := arn[13:]
-	bucketAndFile := strings.SplitN(arnBucketFile, "/", 2)
+	s3ARN, err := arn.Parse(userInput)
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Could not parse arn: %s\n", userInput)
+	}
+	bucketAndFile := strings.SplitN(s3ARN.Resource, "/", 2)
 	if len(bucketAndFile) != 2 {
-		logrus.Fatalf("[FluentBit Init Process] Unrecognizable arn: %s\n", arn)
+		logrus.Fatalf("[FluentBit Init Process] Could not parse arn: %s\n", userInput)
 	}
 
 	bucketName := bucketAndFile[0]


init/fluent_bit_init_process.go
```diff
@@ -195,17 +195,20 @@ func processConfigFile(path string) {
...
 	if !s3ClientCreated {
 		createS3Client()
 	}
 
 	// e.g. "arn:aws:s3:::user-bucket/s3_parser.conf"
-	arnBucketFile := arn[13:]
-	bucketAndFile := strings.SplitN(arnBucketFile, "/", 2)
+	s3ARN, err := arn.Parse(userInput)
+	if err != nil {
+		logrus.Fatalf("[FluentBit Init Process] Unrecognizable arn: %s\n", userInput)
```
ShelbyZ: Should it indicate that this is the s3 arn, similar to the task arn failure?

ShelbyZ (approved):

init/fluent_bit_init_process.go
PettitWesley: It'll print the arn tho, so the user will see its the S3 arn they passed in?

PettitWesley: I think I'll change "unrecognizable" to "failure to parse" or something like that

PettitWesley:
I need to test this with the init integ test before we can merge and release it.

dalgibbard:
> I need to test this with the init integ test before we can merge and release it.

Have tested it on my previously broken deployment in GovCloud and can confirm this is working as expected now 👍

PettitWesley:
So I am struggling to test this myself due issues that I don't fully understand with working in gov cloud. 

I ran the integ test we have with an S3 bucket in the main partition, it worked, so no regression. I then tried running the init with gov cloud credentials and I'm getting:

```
FATA[0000] [FluentBit Init Process] Cannot get bucket region of *****-gov-cloud-test-bucket + dummy-s3-output.conf, you must be the bucket owner to implement this operation
```

It got my bucket name and key correct so the parsing definitely does work.

dalgibbard:
> get bucket region of *****-gov-cloud-test-bucket

I noticed that a lot of the tests have the region hardcoded to ```us-west-2``` - you may need to rewrite those?
```
cd integ/
for fname in `grep -rl "us-west-2" ./*`; do sed -i 's/us-west-2/us-gov-east-1/g' ${fname}; done
```

I also noticed that the tests are using legacy ```docker-compose``` which is deprecated in favour of ```docker compose``` (creating ```/usr/local/bin/docker-compose``` containing ```docker compose $@``` was enough to trick it though)

Other than that, I was having trouble with it expecting various values which I haven't seen documented, like S3_BUCKET_NAME, KINESIS_STREAM, and probably something around S3 actions (causing errors of: ```caused by: failed to perform batch operation on "" to "":```)
I gave up at this point, but seems like some extra docs/configurability around this might be handy too 😅  (at least some notice as to which envvars are expected perhaps).


That said; i'm actively using this in GovCloud right now, so i'm happy that the changes are working anyway :D

PR #621, opened 2023-03-31
Sync Changes for CDK v2

ShelbyZ:
*Description of changes:*

- Move python value to 3.x to support existing CDK v1 sync tasks (only support up to 3.9) and CDK v2 sync tasks (support starts with 3.10)
- Avoids needing to update all sync tasks to CDK v2 immediately

This was tested in an AWS account against CDK v1 and v2 deployments without issues.

### DIFF
--- a/buildspec_sync.yml
+++ b/buildspec_sync.yml
@@ -2,7 +2,7 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      python: 3.7
+      python: 3.x
   pre_build:
     commands:
       - echo Sync latest image from Amazon ECR Public Gallery
--- a/buildspec_sync.yml
+++ b/buildspec_sync.yml
@@ -2,7 +2,7 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      python: 3.x
+      python: 3.10
   pre_build:
     commands:
       - echo Sync latest image from Amazon ECR Public Gallery


matthewfala (approved):
Looks good!

buildspec_sync.yml
```diff
@@ -2,7 +2,7 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      python: 3.7
+      python: 3.x
```
PettitWesley: I think this doesn't work in gov cloud or china... I think we've tried this before... unless something has changed... let's check the git blame

PettitWesley: See here: https://github.com/aws/aws-for-fluent-bit/commit/edbbb2c92f1bb495da45f57d538c84d6858e1ae8

PettitWesley: I doubt they have changed it

PettitWesley (changes requested):
https://github.com/aws/aws-for-fluent-bit/commit/edbbb2c92f1bb495da45f57d538c84d6858e1ae8

I am worried this still won't work in gov cloud

PettitWesley:
> Avoids needing to update all sync tasks to CDK v2 immediately

Why not update them all in one deployment??

ShelbyZ:
> [edbbb2c](https://github.com/aws/aws-for-fluent-bit/commit/edbbb2c92f1bb495da45f57d538c84d6858e1ae8)
> 
> I am worried this still won't owrk in gov cloud

after checking gov-cloud on both cdk v1 and v2 it does look like python: 3.x is not supported as a feature for either

ShelbyZ:
When this change is approved, we will need to update all sync tasks from CDK v1 to v2. CDK v1 will not be able to find an image that satisfies 3.10 (3.9 is the latest) and will cause the tasks to fail pulling new images.

buildspec_sync.yml
```diff
@@ -2,7 +2,7 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      python: 3.7
+      python: 3.10
```
PettitWesley, line 5: python 3.9 instead

PettitWesley:
We should check of python 3.10 now exists in codebuild gov cloud

PettitWesley (approved):
We realized we don't use CDK in sync tasks, and mostly do not use python, so we can just merge this.

matthewfala:
3.10 will NOT work in existing codebuild projects due to python 3.10 support not found in codebuild image STANDARD_2.

3.10 will require an infrastructure update across all sync accounts.

matthewfala:
Closing as this PR will break the existing sync tasks without an infra update.

PR #638, opened 2023-04-14
debugging: add net.dns.resolver

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -389,7 +389,24 @@ And here is an example with `UDP`:
     net.dns.mode UDP
 ```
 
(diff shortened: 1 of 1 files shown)


matthewfala (approved):
Thank you!

matthewfala:
Should we check with Leonardo if we should recommend this?

matthewfala:
I checked with Leonardo. He mentioned it is safe to recommend legacy DNS network provider.
He says that for the new async one: The resolver is limited to one connection attempt which does not allow for DNS fall backs to be used.

He also recommends increasing workers on the output to 2 or 3 if possible (maybe mention that this will not work for s3 or cloudwatch)  You might want to add this to the guide text.

PR #643, opened 2023-04-25
README: make our philosophy on using semver more clear

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The version of the AWS for Fluent Bit image is not linked to the version of Flue
 
 **What does the version number signify?**
 
-We use the standard `major.minor.patch` versioning scheme for our image, AKA Semantic Versioning. The initial release with this versioning scheme is `2.0.0`. An update to the patch version indicates backwards compatible bug fixes, a minor version change indicates new backwards compatible functionality, and a major version change indicates backwards incompatible changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,13 @@ The AWS for Fluent Bit image includes the following contents:
 
 A change in any one of these pieces would lead to a change in our version number.
 
+**Are there edge cases to the rules on breaking backwards compatibility?**
```
PettitWesley, line 50: If anyone disagrees with adding these notes, let's discuss. We could remove them. We definitely need to be more clear that there are ways in which we do sometimes break backwards compatibiity. One  other option is just to be more vague and not ot make any guarantee.

README.md
```diff
@@ -47,6 +47,13 @@ The AWS for Fluent Bit image includes the following contents:
...
+
+Another edge case to the above rules are bug fixes that require removing a change. We have and will continue to occasionally remove new changes in a patch version if they were found to be buggy. We do this to unblock customers who do not depend on the recent change. Please always check our release notes for the changes in a specific version. A past example of a patch release that removed something is [2.31.4](https://github.com/aws/aws-for-fluent-bit/releases/tag/v2.31.4). A prior release had fixed how S3 handles the timestamps in S3 keys and the `Retry_Limit` configuration option. Those changes were considered to be bug fixes. However, they introduced instability so we subsequently removed them in a patch.
```
matthewfala, line 54: This would include the s3 no preserve data ordering changes away from default right?

matthewfala (approved):

README.md
PettitWesley: That only went in FLB 2.0=> change that field to default. And so FLB 2.0 would be 3.0 in our distro I think, so this is a major version change.

PR #650, opened 2023-05-13
tools: small script to write bytes to a file that are not valid utf-8

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/troubleshooting/tools/invalid-utf-8/invalid.py
+++ b/troubleshooting/tools/invalid-utf-8/invalid.py
@@ -1,0 +1,34 @@
+# small script that writes byte values to a file
+# which if interpreted as utf-8 will be invalid chars
+# can test FLB processing of invalid unicode
+
+myfile = open(sys.argv[1], 'ab')
+
+# from https://github.com/fluent/fluent-bit/pull/4297
+invalid = b'\xEF\xBF\x00'
+
+# these were copied from invalid utf-8 examples online
+# https://stackoverflow.com/questions/1301402/example-invalid-utf8-string
+other_invalid_byte_sequence=b'\xC2\xC2\xC0\xC1'
+invalid_examples = [
+    b'\xc3\x28',
+    b'\xa0\xa1',
+    b'\xe2\x28\xa1',
+    b'\xe2\x82\x28',
+    b'\xf0\x28\x8c\xbc',
+    b'\xf0\x90\x28\xbc',
+    b'\xf0\x28\x8c\x28',
+]
+
+
+myfile.write(bytes('some valid line\n', 'utf-8'))
+
+for i in range(255):
+   a_byte = i.to_bytes(1, 'big')
+   myfile.write(a_byte)
+   myfile.write(invalid)
+   myfile.write(other_invalid_byte_sequence)
+   for example in invalid_examples:
+      myfile.write(example)
+
+myfile.write(bytes('some valid line\n', 'utf-8'))


troubleshooting/tools/invalid-utf-8/invalid.py
```diff
@@ -0,0 +1,33 @@
+# small script that writes byte values to a file
+# which if interpreted as utf-8 will be invalid chars
+# can test FLB processing of invalid unicode
+
+myfile = open('log.log', 'ab')
+
+# from https://github.com/fluent/fluent-bit/pull/4297
+invalid = b'\xEF\xBF\x00'
+
+# these were copied from invalid utf-8 examples online
```
weijianz-opc, line 10: should we add source link ?

weijianz-opc: like the line 7

troubleshooting/tools/invalid-utf-8/invalid.py
```diff
@@ -0,0 +1,33 @@
+# small script that writes byte values to a file
+# which if interpreted as utf-8 will be invalid chars
+# can test FLB processing of invalid unicode
+
+myfile = open('log.log', 'ab')
```
weijianz-opc: Consider rename it to invalid_output_file

troubleshooting/tools/invalid-utf-8/invalid.py
```diff
@@ -0,0 +1,33 @@
...
+other_invalid_byte_sequence=b'\xC2\xC2\xC0\xC1'
+invalid_examples = [
+    b'\xc3\x28',
+    b'\xa0\xa1',
+    b'\xe2\x28\xa1',
+    b'\xe2\x82\x28',
+    b'\xf0\x28\x8c\xbc',
+    b'\xf0\x90\x28\xbc',
+    b'\xf0\x28\x8c\x28',
+]
```
zwj102030, line 20: Why don't we put them into one array ?? Why need to one array and two single `string` ? so line 26 - line 27 can be dropped.

troubleshooting/tools/invalid-utf-8/invalid.py
```diff
@@ -0,0 +1,33 @@
...
+
+myfile.write(bytes('some valid line\n', 'utf-8'))
+
+for i in range(255):
+   a_byte = i.to_bytes(1, 'big')
+   myfile.write(a_byte)
+   myfile.write(invalid)
+   myfile.write(other_invalid_byte_sequence)
+   for example in invalid_examples:
+      myfile.write(example)
```
weijianz-opc, line 32: Why can't we do 
```
myfile .write(str.join('', invalid_examples))
```

troubleshooting/tools/invalid-utf-8/invalid.py
PettitWesley: added: https://stackoverflow.com/questions/1301402/example-invalid-utf8-string

troubleshooting/tools/invalid-utf-8/invalid.py
PettitWesley: I think this might do some string conversion on them first, and since they are invalid utf 8 and python uses utf-8, I think that might change them. I think I wnat to write them as bytes. 

This also won't work since they are byte type.

PettitWesley:
@zwj102030 addressed your comments, will you approve now?

troubleshooting/tools/invalid-utf-8/invalid.py
PettitWesley: I'll make it configurable via the first arg on command line

PR #657, opened 2023-05-18
load_tests: increase sleep time before validation and reduce GetLogEvents TPS

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/load_tests/load_test.py
+++ b/load_tests/load_test.py
@@ -16,7 +16,7 @@ TESTING_RESOURCES_STACK_NAME = os.environ['TESTING_RESOURCES_STACK_NAME']
 PREFIX = os.environ['PREFIX']
 EKS_CLUSTER_NAME = os.environ['EKS_CLUSTER_NAME']
 LOGGER_RUN_TIME_IN_SECOND = 600
-BUFFER_TIME_IN_SECOND = 300
+BUFFER_TIME_IN_SECOND = 1000
 NUM_OF_EKS_NODES = 4
 if OUTPUT_PLUGIN == 'cloudwatch':
     THROUGHPUT_LIST = json.loads(os.environ['CW_THROUGHPUT_LIST'])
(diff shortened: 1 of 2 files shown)


load_tests/load_test.py
```diff
@@ -16,7 +16,7 @@
 PREFIX = os.environ['PREFIX']
 EKS_CLUSTER_NAME = os.environ['EKS_CLUSTER_NAME']
 LOGGER_RUN_TIME_IN_SECOND = 600
-BUFFER_TIME_IN_SECOND = 300
+BUFFER_TIME_IN_SECOND = 1000
```
weijianz-opc, line 19: Why we increase the buffer time ?

load_tests/load_test.py
```diff
@@ -54,8 +54,8 @@ def get_log_delay(log_delay_epoch_time):
 # Set buffer for waiting all logs sent to destinations (~5min)
 def set_buffer(stop_epoch_time):
     curr_epoch_time = time.time()
-    if curr_epoch_time-stop_epoch_time < BUFFER_TIME_IN_SECOND:
-        time.sleep(int(BUFFER_TIME_IN_SECOND-curr_epoch_time+stop_epoch_time))
+    if (curr_epoch_time - stop_epoch_time) < BUFFER_TIME_IN_SECOND:
```
weijianz-opc, line 57: do we need brackets ?

load_tests/load_test.py
PettitWesley: See the comment on the other file. This comes from recent testing, if you call GetLogEvents to quickly after a put, you get inconsistent results. The longer you wait, the more likely eventual consistency will not hurt you. 

If you call at too high of a TPS, you also can get inconsistent results.

load_tests/validation/validate.go
```diff
@@ -229,6 +229,14 @@ func validate_cloudwatch(cwClient *cloudwatchlogs.CloudWatchLogs, logGroup strin
 			}
 		}
 
+		/*
+		 * In testing we have found that CW GetLogEvents results are highly inconsistent
+		 * Re-running validation long after tests shows that fewer events were lost than
+		 * first calculated. So we sleep between calls to ensure we never exceed 1 TPS
+		 * load_test.py also has a sleep before validation runs.
+		 */
+		time.Sleep(1 * time.Second)
```
weijianz-opc, line 238: Non block : it will be better have a constant , like` OneSec =  1 * time.Second`

load_tests/load_test.py
PettitWesley: I checked and the math does work

PettitWesley: By brackets do you mean `{`?? Python doesn't use those

lubingfeng (approved):

PR #665, opened 2023-05-31
scripts: add tool to get athena query to check downloads

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/scripts/image_downloads_query.sh
+++ b/scripts/image_downloads_query.sh
@@ -1,0 +1,36 @@
+VERSION=$1
+
+image_shas=$(aws ecr describe-images --registry-id 906394416424 --repository-name aws-for-fluent-bit --region us-west-2 --image-ids imageTag=arm64-${VERSION} imageTag=amd64-${VERSION} imageTag=init-arm64-${VERSION} imageTag=init-amd64-${VERSION} imageTag=${VERSION}-ltsc2019 imageTag=${VERSION}-ltsc2022 | jq -r '.imageDetails[] | objects | .imageDigest')
+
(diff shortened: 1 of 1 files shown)


scripts/image_downloads_query.sh
```diff
@@ -0,0 +1,34 @@
...
+image_sha_array=($image_shas)
+
+echo "Paste below query into Athena console and replace eventTime with desired data range"
+echo "-------"
+echo ""
+echo "
+SELECT COUNT(*) as call_count
+FROM cloudtrail_fluentbitimage
+WHERE
+    eventsource = 'ecr.amazonaws.com' AND
```
lubingfeng, line 16: Is this just querying the public ecr? How about private ecrs?

lubingfeng: We should call out this is for classic regions only.

PettitWesley: I have added:

> echo "This can be used to obtain 'private' ECR download count"
echo "In non-classic regions the table name is 'cloudtrail_fluentbit'"

We call out what you desire in our runbook. The above info in the script is accurate. The query printed is for classic but can be used in other regions by changing the table name as noted.

lubingfeng: We may want to make the script messages explicit by saying: 
The query is for classic region but can be used in other regions by changing the table name.

PettitWesley: Updated to your exact wording

lubingfeng (approved):

PR #630, opened 2023-04-10
Add log loss benchmark framework

PettitWesley:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #670, opened 2023-06-05
log-loss-framework: all test materials developed for AWSLogs benchmark

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/ec2_launcher.sh
+++ b/troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/ec2_launcher.sh
@@ -1,0 +1,57 @@
+export TASK_DEF_FILE=task-def.json
+export RUN_TASK_FILE=ec2-run-task.json
+export TOTAL_SIZE_IN_MB="1000"
+# throughput unit is KB/s
+export THROUGHPUTS="
+1000
+2000
+3000
+4000
+4500
+5000
+6000
+7000
+8000
+9000
+10000
+12000
+"
+# buffer size unit is megabyte
+export BUFFER_SIZES="
+1m
+2m
+4m
+6m
+8m
+12m
+"
+
+export LOG_SIZES="
+1
+250
+"
+
+for SIZE_IN_KB in ${LOG_SIZES}; do
+    export  SIZE_IN_KB=${SIZE_IN_KB}
+    for MAX_BUFFER_SIZE in ${BUFFER_SIZES}; do
+        export MAX_BUFFER_SIZE=${MAX_BUFFER_SIZE}
+        export OUTPUT_FILE=ec2-"${MAX_BUFFER_SIZE}"-buffer-"${SIZE_IN_KB}"-log-size-output.txt
+        export TASK_DEF_OUTPUT_FILE=ec2-"${MAX_BUFFER_SIZE}"-buffer-"${SIZE_IN_KB}"-log-size-taskdefs.txt
+        for test_throughput in ${THROUGHPUTS}; do
+            export THROUGHPUT=$test_throughput
+            export TEST_NAME="$THROUGHPUT"KB_"${SIZE_IN_KB}"KB_AWSLOGS_PATCHED_"${MAX_BUFFER_SIZE}"_BUFFER_EC2
+            TASK_DEF=$(envsubst < "$TASK_DEF_FILE")
+            echo "$TASK_DEF" >> $TASK_DEF_OUTPUT_FILE
+            TASK_DEF_ARN=$(aws ecs register-task-definition --region us-west-2 --cli-input-json  "$TASK_DEF"  | jq '.taskDefinition.taskDefinitionArn')
+            export TASK_DEF_ARN=$(echo "$TASK_DEF_ARN" | tr -d '"')
+            echo "Running $THROUGHPUT $TEST_NAME $TASK_DEF_ARN"
+            RUN_TASK=$(envsubst < "$RUN_TASK_FILE")
+            aws ecs --region us-west-2 run-task --cli-input-json "$RUN_TASK" >> "$OUTPUT_FILE"
+            echo "Started 10 tasks"
+            sleep 100
+            aws ecs --region us-west-2 run-task --cli-input-json "$RUN_TASK" >> "$OUTPUT_FILE"
+            echo "Started 10 tasks"
+            sleep 1000
+        done
+    done
+done
--- a/troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
+++ b/troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
@@ -1,0 +1,92 @@
+{
+    "family": "awslogs-benchmarking-fargate-${MAX_BUFFER_SIZE}-buffer-automated-250K-log",
+    "taskRoleArn": "arn:aws:iam::144718711470:role/ecs_task_role",
+    "executionRoleArn": "arn:aws:iam::144718711470:role/ecsTaskExecutionRole",
+    "networkMode": "awsvpc",
+    "cpu": "1024",
+    "memory": "2048",
+    "containerDefinitions": [
+      {
+        "essential": false,
+        "image": "144718711470.dkr.ecr.us-west-2.amazonaws.com/log-loss-benchmark-logger:latest",
+        "name": "logger",
+        "logConfiguration": {
+          "logDriver": "awslogs",
+          "options": {
+            "awslogs-group": "awslogs-benchmarking-output",
+            "awslogs-region": "${REGION}",
+            "awslogs-create-group": "true",
+            "awslogs-stream-prefix": "pre-testing-bug",
+            "mode": "non-blocking",
+            "max-buffer-size": "${MAX_BUFFER_SIZE}"
+          }
+        },
+        "environment": [
+          {
+            "name": "SIZE_IN_KB",
+            "value": "${SIZE_IN_KB}"
+          },
+          {
+            "name": "TOTAL_SIZE_IN_MB",
+            "value": "${TOTAL_SIZE_IN_MB}"
+          },
+          {
+            "name": "THROUGHPUT_IN_KB",
+            "value": "${THROUGHPUT}"
+          },
+          {
+            "name": "CYCLE_TIME_IN_SECONDS",
+            "value": "1"
+          }
+        ],
+        "memoryReservation": 50
+      },
+      {
+        "essential": true,
+        "image": "144718711470.dkr.ecr.us-west-2.amazonaws.com/cw-log-loss-framework-validator:latest",
+        "name": "validator",
+        "logConfiguration": {
+          "logDriver": "awslogs",
+          "options": {
+            "awslogs-group": "awslogs-benchmarking-results",
+            "awslogs-region": "${REGION}",
+            "awslogs-create-group": "true",
+            "awslogs-stream-prefix": "pre-testing-bug"
+          }
+        },
+        "dependsOn": [
+          {
+            "containerName": "logger",
+            "condition": "SUCCESS"
+          }
+        ],
+        "memoryReservation": 100,
+        "environment": [
+          {
+            "name": "SIZE_IN_KB",
+            "value": "${SIZE_IN_KB}"
+          },
+          {
+            "name": "TOTAL_SIZE_IN_MB",
+            "value": "${TOTAL_SIZE_IN_MB}"
+          },
+          {
+            "name": "THROUGHPUT_IN_KB",
+            "value": "${THROUGHPUT}"
+          },
+          {
+            "name": "TEST_NAME",
+            "value": "${TEST_NAME}"
+          },
+          {
+            "name": "CW_LOG_GROUP_NAME",
+            "value": "awslogs-benchmarking-output"
+          },
+          {
+            "name": "CW_LOG_STREAM_PREFIX",
+            "value": "pre-testing-bug/logger/"
+          }
+        ]
+      }
+    ]
+  }
--- a/troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate_launcher.sh
+++ b/troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate_launcher.sh
@@ -1,0 +1,59 @@
+export TASK_DEF_FILE=fargate-task.json
+export RUN_TASK_FILE=fargate-run-task.json
+export TOTAL_SIZE_IN_MB="1000"
+export REGION='us-west-2'
+# throughput unit is KB/s
+export THROUGHPUTS="
+1000
+2000
+3000
+4000
+4500
+5000
+6000
+7000
+8000
+9000
+10000
+12000
+"
+# buffer size unit is megabyte
+export BUFFER_SIZES="
+1m
+2m
+4m
+6m
+8m
+12m
+"
+
+export LOG_SIZES="
+1
+250
+"
+
+for SIZE_IN_KB in ${LOG_SIZES}; do
+    export  SIZE_IN_KB=${SIZE_IN_KB}
+    for MAX_BUFFER_SIZE in ${BUFFER_SIZES}; do
+        export MAX_BUFFER_SIZE=${MAX_BUFFER_SIZE}
+        export OUTPUT_FILE=fargate-"${MAX_BUFFER_SIZE}"-buffer-"${SIZE_IN_KB}"-log-size-output.txt
+        export TASK_DEF_OUTPUT_FILE=fargate-"${MAX_BUFFER_SIZE}"-buffer-"${SIZE_IN_KB}"-log-size-taskdefs.txt
+        for test_throughput in ${THROUGHPUTS}; do
+            export THROUGHPUT=$test_throughput
+            export TEST_NAME="$THROUGHPUT"KB_"${SIZE_IN_KB}"KB_AWSLOGS_"${MAX_BUFFER_SIZE}"_BUFFER_FARGATE_AUTO
+            TASK_DEF=$(envsubst < "$TASK_DEF_FILE")
+            echo "$TASK_DEF" >> $TASK_DEF_OUTPUT_FILE
+            TASK_DEF_ARN=$(aws ecs register-task-definition --region ${REGION} --cli-input-json  "$TASK_DEF"  | jq '.taskDefinition.taskDefinitionArn')
+            export TASK_DEF_ARN=$(echo "$TASK_DEF_ARN" | tr -d '"')
+            echo "Running $THROUGHPUT $TEST_NAME $TASK_DEF_ARN"
+            RUN_TASK=$(envsubst < "$RUN_TASK_FILE")
+            aws ecs --region ${REGION} run-task --cli-input-json "$RUN_TASK" >> "$OUTPUT_FILE"
+            echo "Started 10 tasks"
+            sleep 100
+            aws ecs --region ${REGION} run-task --cli-input-json "$RUN_TASK" >> "$OUTPUT_FILE"
+            echo "Started 10 tasks"
+            # can run more concurrent fargate tasks
+            sleep 500
+        done
+    done
+done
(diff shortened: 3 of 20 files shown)


PettitWesley:
Original review and comments here were addressed: https://github.com/aws/aws-for-fluent-bit/pull/630

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/ec2_launcher.sh
```diff
@@ -0,0 +1,55 @@
...
+5000
+6000
+7000
+8000
+9000
+10000
+12000
+"
+export BUFFER_SIZES="
+1m
```
weijianz-opc, line 21: it better have a comment to tell what is the unit for the buffer size , 1MB ?

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
```diff
@@ -0,0 +1,92 @@
...
+					"value": "${TOTAL_SIZE_IN_MB}"
+				},
+                {
+					"name": "THROUGHPUT_IN_KB",
+					"value": "${THROUGHPUT}"
+				},
+                {
+					"name": "CYCLE_TIME_IN_SECONDS",
+					"value": "1"
+				}
```
weijianz-opc: Should we fix the indention  ??

weijianz-opc: Also, why others are env variable but not this `"1"`

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
```diff
@@ -0,0 +1,92 @@
...
+				},
+                {
+					"name": "TEST_NAME",
+					"value": "${TEST_NAME}"
+				},
+                {
+					"name": "CW_LOG_GROUP_NAME",
+					"value": "awslogs-benchmarking-output"
+				},
+                {
```
weijianz-opc: Same indention

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate_launcher.sh
```diff
@@ -0,0 +1,57 @@
...
+12000
+"
+export BUFFER_SIZES="
+1m
+2m
+4m
+6m
+8m
+12m
+"
```
weijianz-opc: Can't we move those into util class, why it needs to be redefined here ??  Since `ecs_launcher.sh` has same variables

weijianz-opc:
This seems a fairly large PR. It will be better to break them into smaller ones.

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
PettitWesley: The CYCLE_TIME_IN_SECONDS doesn't change in any of the tests, so I didn't add it as an env

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate_launcher.sh
PettitWesley: this is a simple quick script to make launching easy for the AWSLogs project... it may never be used again... so I think its not worth it

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/ec2_launcher.sh
PettitWesley: added

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
PettitWesley: fixed indentation

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate-task.json
PettitWesley: fixed indentation

troubleshooting/tools/log-loss-test-framework/validator/validate.go
```diff
@@ -0,0 +1,383 @@
...
+	if (totalInputRecord - uniqueRecordFound) != missingRecordCount {
+		fmt.Printf("%s: VALIDATION ERROR: (totalInputRecord - uniqueRecordFound) = %d, missingRecordCount = %d\n", 
+					testName, totalInputRecord - uniqueRecordFound, missingRecordCount)
+	}
+
+	// if there are many test cases to run
+	// perform CW insights query for test name
+	// output is comma delimited, output insights query result as CSV
+	// simple python code can parse the CSV
+	fmt.Printf("%s %s - %s, percent lost, %d, number_lost, %d, total_input_record, %d, duplicates, %d, group=%s stream=%s TOTAL_SIZE_IN_MB=%s, SIZE_IN_KB=%s, THROUGHPUT_IN_KB=%s, %s, %s, %s, %s, %s, first_lost=%d, last_lost=%d",
```
matthewfala, line 361: I really like how task metadata is included along with all of the logging statistics. It allows all the data to be dumped into one CW stream making the data easily queryable rather than having a separate stream for each test/result set.

troubleshooting/tools/log-loss-test-framework/validator/validate.go
```diff
@@ -0,0 +1,383 @@
...
+				if makeHistogram {
+					bucket := (lostID - idCounterBase) / bucketSize
+					missingList := missingRecords[bucket]
+					missingList = append(missingList, lostID)
+					missingRecords[bucket] = missingList
+				}
+			}
+		}
+	}
+	// fmt.Printf("\n")
```
matthewfala, line 343: Consider removing

troubleshooting/tools/log-loss-test-framework/validator/validate.go
```diff
@@ -0,0 +1,383 @@
...
+				response, err = cwClient.GetLogEvents(input)
+			} else {
+				return fmt.Errorf("[TEST FAILURE] Error occured to get the log events from log group: %q., %v", logGroup, err), 0, nil
+			}
+		}
+		if makeHistogram != "" {
+			fmt.Printf(".")
+		}
+
+		/* sleep between GetLogEvents calls to proactively reduce TPS against CW frontend */
```
matthewfala, line 260: If you want to be consistent with style, use //

troubleshooting/tools/log-loss-test-framework/validator/validate.go
```diff
@@ -0,0 +1,383 @@
...
+			}
+			fmt.Println(" ")
+		}
+	}
+}
+
+func exitErrorf(msg string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, msg+"\n", args...)
+	os.Exit(1)
+}
```
matthewfala, line 383: The logic of this file makes sense. I like the use of map, and also storing the duplicated log count via the difference between the counted logs and the number of logs appearing in the map. The histogram idea is also nice. Generally the comments in the code file start with both upper and lower case, which to convention it may be better to start all of them with upper or lower, rather than both though this is really minor and not really an essential solution.

troubleshooting/tools/log-loss-test-framework/results/test_run_merge.py
```diff
@@ -0,0 +1,161 @@
...
+        'log_group': log_group,
+        'log_stream': log_stream,
+        'total_mb': total_mb,
+    }
+
+main_set = {}
+
+if len(sys.argv) < 3:
+    # Created due to the large number of test cases that passed after triple+ re-validation
+    print("Usage: merges at least two result files, taking the more successful run when duplicates are found. Can specify any number of files in any order.")
```
matthewfala, line 118: I don't see any mention of duplicates in the code. Looks like it's just taking the data of the sample per task definition with the lowest number_lost metric

troubleshooting/tools/log-loss-test-framework/results/test_run_merge.py
```diff
@@ -0,0 +1,161 @@
...
+                continue
+            
+            task_id = data['task_id']
+            test_str = ", ".join(test_run)
+            data['str'] = test_str
+        
+            if task_id in main_set:
+                re_validated += 1
+                existing = main_set[task_id]
+                if data['number_lost'] < existing['number_lost']:
```
matthewfala, line 147: Is this a good idea? I'm wondering if a better solution is median -- that is: to order the samples from worst to best and choose the middle sample to exclude outliers. That way we don't select the outlier best result.

troubleshooting/tools/log-loss-test-framework/results/test_run_filter.py
```diff
@@ -0,0 +1,145 @@
...
+            log_group = kv[1]
+        if kv[0] == 'TOTAL_SIZE_IN_MB':
+            total_mb = kv[1]
+    log_stream = os.environ.get('CW_LOG_STREAM_PREFIX', 'pre-testing-bug/logger/') + task_id
+
+    # use task ID to guard against duplicate test results pasted into the output due to human error
+    if task_ids_found.get(task_id):
+        print("ERROR: duplicate test run "+ test_name)
+        return None
+    task_ids_found[task_id] = True
```
matthewfala, line 103: Seems like this file, except this portion and the main code section below is copy pasted from other files. Consider creating a test_run_parsers file that includes common validation and test run logic to prevent a headache making changes to all the files later.

troubleshooting/tools/log-loss-test-framework/results/test_run_filter.py
```diff
@@ -0,0 +1,145 @@
...
+            continue
+
+        data = process_test_run(test_run)
+        if data == None:
+            print("INVALID DATA: " + ", ".join(test_run))
+            continue
+
+        # manually put in critera to separate tests
+        if not data['success'] and data['throughput'] > 10000 and data['throughput'] < 25000 and data['buffer'] >= 15:
+            print(", ".join(test_run))
```
matthewfala, line 143: Should there be a new line here since adding a comma may put everything on one very long line, unless I'm missing that test_run has a new line.

troubleshooting/tools/log-loss-test-framework/results/test_run_filter.py
```diff
@@ -0,0 +1,145 @@
...
+
+
```
matthewfala, line 145: nit: Couple extra white spaces.

troubleshooting/tools/log-loss-test-framework/results/test_run_diff.py
```diff
@@ -0,0 +1,211 @@
...
+
+# arn:aws:logs:region:account-id:log-group:log_group_name:log-stream:log-stream-name
+
+for task_id in sorted(first_set.keys()):
+    if task_id in second_set:
+        first = first_set[task_id]
+        second = second_set[task_id]
+        if first['number_lost'] != second['number_lost']:
+            re_validated += 1
+            diff = first['number_lost'] - second['number_lost']
```
matthewfala, line 190: I think diff usually does (second - first). The convention here may be different though

troubleshooting/tools/log-loss-test-framework/results/test_run_diff.py
```diff
@@ -0,0 +1,211 @@
...
+            continue
+            
+        task_id = data['task_id']
+        test_str = ", ".join(test_run)
+        data['str'] = test_str
+        
+        second_set[task_id] = data
+
+
+
```
matthewfala, line 169: Nit: inconsistent new line, should be 2

troubleshooting/tools/log-loss-test-framework/results/test_run_checker.py
```diff
@@ -0,0 +1,188 @@
...
+    cmd_env['ECS_TASK_DEFINITION'] = test_run['task_def']
+    cmd_env['ECS_CLUSTER'] = test_run['cluster']
+    cmd_env['ECS_TASK_ID'] = test_run['task_id']
+    cmd_env['ECS_LAUNCH_TYPE'] = test_run['launch_type']
+    cmd_env['HISTOGRAM'] = 'On'
+
+    subprocess.run([sys.argv[1]], capture_output=False, env=cmd_env)
+    # result = subprocess.run([sys.argv[1]], capture_output=True, env=cmd_env, encoding='utf-8')
+    # for line in result.stdout.split('\n'):
+    #     print(line)
```
matthewfala, line 158: Where does the data go? As far as I can tell, validator.go sends everything to stdout (or err). I would expect the result to be stored. What's going on here.

troubleshooting/tools/log-loss-test-framework/results/analyze.py
```diff
@@ -0,0 +1,395 @@
...
+
+        if max_loss == 0:
+            table_summary_max = f'{metric_max}'
+        else:
+            table_summary_max = f'{metric_max} {max_loss:.2f}%'
+
+        metric_success = '🚨'
+        if successful_runs == total_runs:
+            metric_success = '✅'
+        elif (successful_runs / total_runs) > 0.9 and (total_runs - successful_runs) == 1:
```
matthewfala, line 253: I suppose the assumption here is that one time errors/transient errors are less worrysome.

troubleshooting/tools/log-loss-test-framework/results/analyze.py
```diff
@@ -0,0 +1,395 @@
...
+        summary = f'{test_case}: {success} {successful_runs}/{total_runs} Loss % metrics: AVG: {avg_loss}, MAX: {max_loss}, MIN: {min_loss}, STDEV: {stdev_loss}'
+        results.append(summary)
+
+    # sort so that data is ordered with buffer sizes together
+    results.sort()
+    for result in results:
+        print(result)
+
+    # All results compiled into one table:
+    results_table_with_dimensions(all_test_runs, {})
```
matthewfala, line 383: Are these tables untitled but listed one after the other if there are datapoints with these keyvalue entries?
May be helpful to have titles.

troubleshooting/tools/log-loss-test-framework/logger/log_generator.c
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    loggingEnd = timeInMilliseconds();
+    timeDiff = loggingEnd - loggingStart;
+    double timeDiffD = (double) timeDiff;
+    double sizeSent = (double) messagesSent * sizeInKb;
+    double realThroughput = sizeSent / (timeDiff / 1000);
+    printf("real runtime: %lldms\n", timeDiff);
+    printf("emitted: %dkb\n", messagesSent * sizeInKb);
```
matthewfala, line 153: I think this should be upper case KB? The reader could think this means kilobits:
```
KB stands for "Kilobytes" which is a storage measure.

kb, on the other hand, stands for "kilobits" which is a common measure for internet speed.
```
https://stackoverflow.com/questions/62489990/what-is-the-value-of-uppercase-kb-in-computer-science-what-is-difference-betw

troubleshooting/tools/log-loss-test-framework/logger/log_generator.c
```diff
@@ -0,0 +1,174 @@
...
+        printf("Error opening file!\n");
+        /* 
+         * if a volume isn't configured for this path it might fail 
+         * so just exit with success. When we run in ECS validator has a SUCCESS
+         * dependency on logger. 
+         */
+        return 0;
+    }
+    fprintf(f, "real runtime: %lldms\n", timeDiff);
+    fprintf(f, "emitted: %dkb\n", messagesSent * sizeInKb);
```
matthewfala, line 169: Same KB

troubleshooting/tools/log-loss-test-framework/logger/log_generator.c
```diff
@@ -0,0 +1,174 @@
...
+    int burstSizeInMB = 0;
+    int burstThroughputInKb = 0;
+    int burstMessagesPerCycle = 0;
+    int totalBurstMessages = 0;
+    int burstDelayInMB = 0;
+    int burstDelayMessages = 0;
+    if (burst_enabled != NULL) {
+        burstSizeInMB = atoi(burst_enabled);
+        burstThroughputInKb = atoi(getenv("BURST_THROUGHPUT_IN_KB"));
+        if ((burstSizeInMB * 2 * 1000) > totalSizeInKb) {
```
matthewfala, line 79: I don't see why this limit needs to be imposed. burstSizeInMB seems like it needs to be less than totalSizeInKb/1000 rather than less than half that.

troubleshooting/tools/log-loss-test-framework/results/loss_histogram.py
```diff
@@ -0,0 +1,188 @@
...
+    key = ""
+    for field in aggregate_on:
+        val = data[field]
+        key += val
+    if data['buffer'] > 1:
+        key += 'non-default-buffer'
+    data['key'] = key
+    return key
+
+
```
matthewfala, line 187: Function redefinition. Also defined above. I'm guessing that this is the one you want to keep? Consider removing one implementation

troubleshooting/tools/log-loss-test-framework/results/loss_histogram.py
```diff
@@ -0,0 +1,188 @@
...
+    current_data_set = []
+    key = None
+    for line in lines:
+        if TEST_SUMMARY_MARKER in line:
+            # get summary
+            index = line.find('last_lost=')
+            if index > 0:
+                index = line.find(':', index)
+                end = index - 1
+                summary = line[:end]
```
matthewfala, line 129: I would expect this to be:
```
end = index
summary = line[:end]
```
Because end is not included. I suppose the last value which is the test case name from the summary section added by analyze.py is really garbage and not parsed out.

troubleshooting/tools/log-loss-test-framework/logger/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM alpine as build-env
```
matthewfala, line 1: Interesting choice. I suppose alpine can be used instead of amazon2 linux as long as we don't distribute.

troubleshooting/tools/log-loss-test-framework/awslogs-benchmarking-scripts/fargate_launcher.sh
```diff
@@ -0,0 +1,59 @@
...
+            TASK_DEF=$(envsubst < "$TASK_DEF_FILE")
+            echo "$TASK_DEF" >> $TASK_DEF_OUTPUT_FILE
+            TASK_DEF_ARN=$(aws ecs register-task-definition --region ${REGION} --cli-input-json  "$TASK_DEF"  | jq '.taskDefinition.taskDefinitionArn')
+            export TASK_DEF_ARN=$(echo "$TASK_DEF_ARN" | tr -d '"')
+            echo "Running $THROUGHPUT $TEST_NAME $TASK_DEF_ARN"
+            RUN_TASK=$(envsubst < "$RUN_TASK_FILE")
+            aws ecs --region ${REGION} run-task --cli-input-json "$RUN_TASK" >> "$OUTPUT_FILE"
+            echo "Started 10 tasks"
+            sleep 100
+            aws ecs --region ${REGION} run-task --cli-input-json "$RUN_TASK" >> "$OUTPUT_FILE"
```
matthewfala, line 53: So 20 tasks are started for each type of test right? So 6*2*12*20 = 2880 tasks. That's a lot of tasks!

matthewfala (commented):
Looks great! Lot's of insightful metrics collected on the tests. The organization of the result data as a csv where each line represents a different test with all metric dimensions is creative and impactful as it allows for drilling down statistics on dimensions like buffer and throughput, and also obtaining aggregate statistics.

Left comments on the portions of code that I felt may need a second look. Also there are some style comments which I stopped adding half way through, since I'm not sure if you are concerned about that with these scripts.

PR #676, opened 2023-06-08
load-test: fix failing tests due to role session timeout, clean up code

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/buildspec_load_test.yml
+++ b/buildspec_load_test.yml
@@ -52,12 +52,6 @@ phases:
       - python -m venv venv
       - source venv/bin/activate
       - pip install -r ./load_tests/requirements.txt
-      # AWS configure sts temp credentials
-      - |
-        CREDS=`aws sts assume-role --role-arn ${LOAD_TEST_CFN_ROLE_ARN} --duration-seconds 900 --role-session-name load-test-cfn`
-        aws configure set aws_access_key_id `echo $CREDS | jq -r .Credentials.AccessKeyId`
-        aws configure set aws_secret_access_key `echo $CREDS | jq -r .Credentials.SecretAccessKey`
-        aws configure set aws_session_token `echo $CREDS | jq -r .Credentials.SessionToken`
       - |
         if [ "${PLATFORM}" == "EKS" ]; then
           aws eks update-kubeconfig --name $EKS_CLUSTER_NAME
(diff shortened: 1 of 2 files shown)


buildspec_load_test.yml
```diff
@@ -52,21 +52,13 @@ phases:
       - python -m venv venv
       - source venv/bin/activate
       - pip install -r ./load_tests/requirements.txt
-      # AWS configure sts temp credentials
```
PettitWesley, line 55: You may wonder, why did he remove this? Is this safe?

In load tests, everything runs via the python script, and as you can see in the other code changes, that code assumes the `LOAD_TEST_CFN_ROLE_ARN` role itself. This means we were previously unnecessarily chaining sessions (buildspec assumes the role, then the python uses that role to assume the same role).

load_tests/load_test.py
```diff
@@ -224,7 +223,13 @@ def run_ecs_tests():
         # Validation input type banner
         print(f'\nTest {input_logger["name"]} to {OUTPUT_PLUGIN} in progress...')
 
+    # wait for tasks and validate
+    for input_logger in INPUT_LOGGERS:
```
PettitWesley, line 227: This change should speed up the tests slightly. If you read the lines above and below here, we first start the tasks for each input (tcp vs stdout) and then validate... but each of our tests uses a unique resource (log stream, firehose, kinesis, etc) for each test... so there's no reason why we can't start all of them first and then loop over them and check for results.

weijianz-opc (approved):
No concern.

load_tests/load_test.py
```diff
@@ -200,7 +200,6 @@ def run_ecs_tests():
         session = get_sts_boto_session()
 
         client = session.client('ecs')
-        waiter = client.get_waiter('tasks_stopped')
         
         processes = []
```
DrewZhang13, line 205: Do we till need this for loop after logic change? I don't understand what's the work for this for-loop.
At least should we remove `processes = []`  out of the for loop?

PettitWesley: oh you're right, thanks, processes needs to move to below.

PettitWesley: https://github.com/aws/aws-for-fluent-bit/pull/677

PettitWesley: This for loop runs the tasks, then 2nd one validates.

PR #678, opened 2023-06-09
load-test: CFN credentials are needed for CDK deploy

PettitWesley:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/buildspec_load_test.yml
+++ b/buildspec_load_test.yml
@@ -56,9 +56,17 @@ phases:
         if [ "${PLATFORM}" == "EKS" ]; then
           aws eks update-kubeconfig --name $EKS_CLUSTER_NAME
         fi
+      - CREDS==$(aws sts assume-role --role-arn $LOAD_TEST_CFN_ROLE_ARN --role-session-name load-test-cfn --duration-seconds 3600)
+      - export CREDS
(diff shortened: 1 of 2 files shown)


buildspec_load_test.yml
```diff
@@ -56,9 +56,17 @@ phases:
         if [ "${PLATFORM}" == "EKS" ]; then
           aws eks update-kubeconfig --name $EKS_CLUSTER_NAME
         fi
+      - CREDS==$(aws sts assume-role --role-arn $LOAD_TEST_CFN_ROLE_ARN --role-session-name load-test-cfn --duration-seconds 3600)
```
PettitWesley, line 59: basically in my previous PR I missed that the python code assumes the CFN role for everything it does, but then also does a `os.system('cdk deploy --require-approval never')` if we're in the cloudwatch test codebuild project... 

which means it needs the creds from the role in the env: https://github.com/aws/aws-for-fluent-bit/blob/mainline/load_tests/load_test.py#L186

PettitWesley: Basically the logic there exists because we need to wait on the CFN stack in the other tests, but create it in one. Its easier to do that in python than in shell.

DrewZhang13 (approved):

buildspec_load_test.yml
PettitWesley: LOL == double :/

PR #681, opened 2023-06-09
load-test: CFN role trust policy requires session name be exactly 'load-test-cfn'

PettitWesley:
I don't remember why we decided this and I think its unnecessary but the load test CFN role has this:

```
"Action": "sts:AssumeRole",
            "Condition": {
                "StringLike": {
                    "sts:RoleSessionName": "load-test-cfn"
                }
            }
```

So my previous change to the role session name to differentiate the python vs buildspec calls doesn't work..


weijianz-opc (approved):
lgtm , can we add the related roleSeesionName change PR into this PR description ?

DrewZhang13 (approved):

PR #682, opened 2023-06-12
Fix load test failures

PettitWesley:
The main cause of the failures seems to be that task def names are re-used across tests. I made some other improvements as well. 

### DIFF
--- a/load_tests/load_test.py
+++ b/load_tests/load_test.py
@@ -80,10 +80,6 @@ def calculate_total_input_number(throughput):
 # 1. Configure task definition for each load test based on existing templates
 # 2. Register generated task definition
 def generate_task_definition(session, throughput, input_logger, s3_fluent_config_arn):
-    if not hasattr(generate_task_definition, "counter"):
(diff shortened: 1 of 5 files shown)


load_tests/load_test.py
```diff
@@ -394,7 +394,7 @@ def parse_json_template(template, dict):
     for key in dict:
             if(key[0] == '$'):
                 data = data.replace(key, dict[key])
-            elif(key == OUTPUT_PLUGIN):
+            else:
```
weijianz-opc, line 397: we mean to make this change??

weijianz-opc (approved):
All good !

DrewZhang13 (approved):
LGTM

load_tests/load_test.py
PettitWesley: from slack:

this one is safe
previously it would relpace the vars if the key was for the. current output plugin
now it just iterates overall of them






it does a find and replace




so if the string isn't in our template, it own't replace it
this old logic was unneeded and wrong

PR #685, opened 2023-06-13
load-test: differentiate task ARNs for each input logger

PettitWesley:
My change last week sped up the tests by starting all tasks first and then checking them. However, I didn't notice that the code doesn't properly distinguish the task arns for each input logger properly. There are two, TCP and standard out. Currently the standard out runs first and then TCP overwrites the task ARN for standard out. 

THis fixes that. 

Tested by commenting out all boto3 code and running locally which ensures my python is at least valid. 

### DIFF
--- a/load_tests/load_test.py
+++ b/load_tests/load_test.py
@@ -212,7 +212,8 @@ def run_ecs_tests():
                     launchType='EC2',
                     taskDefinition=f'{PREFIX}{OUTPUT_PLUGIN}-{throughput}-{input_logger["name"]}'
             )
-            names[f'{OUTPUT_PLUGIN}_{throughput}_task_arn'] = response['tasks'][0]['taskArn']
+            print(f'run_task_response={response}')
+            names[f'{OUTPUT_PLUGIN}_{input_logger["name"]}_{throughput}_task_arn'] = response['tasks'][0]['taskArn']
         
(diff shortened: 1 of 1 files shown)


weijianz-opc (approved):

DrewZhang13 (approved):

Claych (approved):

load_tests/load_test.py
```diff
@@ -212,7 +212,8 @@ def run_ecs_tests():
                     launchType='EC2',
                     taskDefinition=f'{PREFIX}{OUTPUT_PLUGIN}-{throughput}-{input_logger["name"]}'
             )
-            names[f'{OUTPUT_PLUGIN}_{throughput}_task_arn'] = response['tasks'][0]['taskArn']
+            print(f'run_task_response={response}')
+            names[f'{OUTPUT_PLUGIN}_{input_logger["name"]}_{throughput}_task_arn'] = response['tasks'][0]['taskArn']
```
lubingfeng, line 216: Can we add the comments why we need to add input_logger as the key?

lubingfeng: Per discussion, we used to have stdout as the only source of input. Using tcp as input was added this year has triggered the issue. The fix is good. Thanks.

PR #686, opened 2023-06-13
load-test: delete testing data is only needed at end of tests, since each test uses unique resource prefix

PettitWesley:
### Why is this change needed?

Last week I made a change to speed up the tests and fix a timeout that had occurred by starting all tests first, then checking all results. Basically, run all in parallel vs serial. 

However, the code currently deletes testing data in between each test runs, which no longer works since they run in parallel now. 

### When do we need to delete testing data?

There's already a `delete_testing_resources` phase after all testing runs are completed. That's the right place to delete test data.

The previous code to delete between test runs was entirely unnecessary. Each test run uses a unique log stream in CW or unique S3 key in S3. 

### Why is this change safe?

The previous code to delete between test runs was entirely unnecessary. Each test run uses a unique log stream in CW or unique S3 key in S3. 

### How was this change tested?

I used the commit here to check that the python syntax is correct and that the generated task defs are correct: https://github.com/PettitWesley/aws-for-fluent-bit/commit/d5fd7c1f32852411d0a747ada826925fd97eeee1

Since we do not have a test pipeline already available, its not possible to run the load tests prior to PR merge.

This change is needed to unblock the 2.31.12 release which multiple customers are waiting on. 

### DIFF
--- a/load_tests/load_test.py
+++ b/load_tests/load_test.py
@@ -197,9 +197,6 @@ def run_ecs_tests():
 
         client = session.client('ecs')
 
-        # Delete corresponding testing data for a fresh start
-        delete_testing_data(session)
-
         # S3 Fluent Bit extra config data
         s3_fluent_config_arn = publish_fluent_config_s3(session, input_logger)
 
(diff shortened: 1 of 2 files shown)


load_tests/load_test.py
```diff
@@ -411,27 +408,22 @@ def publish_fluent_config_s3(session, input_logger):
     )
     return f'arn:aws:s3:::{bucket_name}/{OUTPUT_PLUGIN}-test/{PLATFORM}/fluent-{input_logger["name"]}.conf'
 
-# The following method is used to clear data between
-# testing batches
+# The following method is used to clear data after all tests run
 def delete_testing_data(session):
-    # All testing data related to the plugin option will be deleted
-    if OUTPUT_PLUGIN == 'cloudwatch':
```
PettitWesley, line 418: the "delete_testing_resources" only runs once, so its no longer necessary to check the plugin here. The delete happens just once at the end of the cloudwatch tests in the main "delete_testing_resources" routine.

load_tests/logger/stdout_logger/fluent.conf
```diff
@@ -0,0 +1,2 @@
+# no content needed for stdout test
```
PettitWesley, line 1: this change is just meant for maintainence to explain why this file is empty

weijianz-opc (approved):

DrewZhang13 (approved):

PR #687, opened 2023-06-14
load-tests: custom task_stopped waiter that logs state + won't fail due to expired session creds

PettitWesley:
### Why is it taking forever to fix the load tests?

I'm sorry. Also I have gotten the firehose and kinesis and S3 tests to pass with my recent changes. The remaining issue is session expiration in CW tests. 

### Why is this changed needed?

The remaining issue is session expiration in CW tests. 

This change does a couple of things:
1. Wait for tasks stopped using creds from the CodeBuild role, not the CFN role. This session should auto-refresh and should not expire. 
2. The custom waiter prints the task status as it checks which will help with debugging. It also doesn't call ECS as frequently which is more frugal (longer sleep time). 
3. Removes the usage of the CFN special role for actually running the tests. This isn't needed if our CodeBuild Role has ECS permissions and IAM:PassRole. I need to separately change our CDK to add perms to the role (which I've added in the account manually to unblock us). This makes us safer than previously because the CFN role has admin level permissions. 

### Testing

I've tested that this python is syntactically correct. Testing that it works is easiest in the actual pipeline. 

### Future work

We probably do not need the CFN role for resource deletion, only creation. I've left its use for deletion for now. 

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/load_tests/load_test.py
+++ b/load_tests/load_test.py
@@ -81,7 +81,7 @@ def calculate_total_input_number(throughput):
 
 # 1. Configure task definition for each load test based on existing templates
 # 2. Register generated task definition
-def generate_task_definition(session, throughput, input_logger, s3_fluent_config_arn):
+def generate_task_definition(throughput, input_logger, s3_fluent_config_arn):
     # Generate configuration information for STD and TCP tests
     std_config      = resource_resolver.get_input_configuration(PLATFORM, resource_resolver.STD_INPUT_PREFIX, throughput)
     custom_config   = resource_resolver.get_input_configuration(PLATFORM, resource_resolver.CUSTOM_INPUT_PREFIX, throughput)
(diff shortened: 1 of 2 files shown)


load_tests/load_test.py
```diff
@@ -181,29 +184,55 @@ def create_testing_resources():
...
+
 # For tests on ECS, we need to:
 #  1. generate and register task definitions based on templates at /load_tests/task_definitons
 #  2. run tasks with different throughput levels for 10 mins
 #  3. wait until tasks completed, set buffer for logs sent to corresponding destinations
 #  4. validate logs and print the result
 def run_ecs_tests():
     ecs_cluster_name = os.environ['ECS_CLUSTER_NAME']
-    names = locals()
+    names = {}
```
PettitWesley, line 222: This existing code is not needed, names is just a dict

load_tests/setup_test_environment.sh
```diff
@@ -14,6 +14,8 @@ stackOutputs=$(aws cloudformation describe-stacks --stack-name ${LOG_STORAGE_STA
 read -r -a outputArray <<< "$stackOutputs"
 export S3_BUCKET_NAME="${outputArray[0]}"
 
+export AWS_DEFAULT_REGION=${AWS_REGION}
```
PettitWesley, line 17: This probably isn't needed but it doesn't hurt. I added it sicne AWS_REGION is set in Code Build but it seems boto3 from local test liked AWS_DEFAULT_REGION

Claych (approved):

PR #689, opened 2023-06-14
load-test: increase waiter sleep time and handle task missing

PettitWesley:
This isn't needed to fix any failure or code bug, it just makes the code I contributed in last PR a little better.

1. Increase sleep wait time so that it prints describe task output less often (I want the full output for debugging)
2. Handle missing task response, which shouldn't happen normally, but did once in my local test, not clear why. 

### DIFF
--- a/load_tests/load_test.py
+++ b/load_tests/load_test.py
@@ -253,6 +253,9 @@ def run_ecs_tests():
         # Validation input type banner
         print(f'\nTest {input_logger["name"]} to {OUTPUT_PLUGIN} in progress...')
 
+    # Tasks need time to run
+    time.sleep(600)
+
     # wait for tasks and validate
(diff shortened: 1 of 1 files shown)


load_tests/load_test.py
```diff
@@ -9,8 +9,9 @@
 from datetime import datetime, timezone
 import create_testing_resources.kinesis_s3_firehose.resource_resolver as resource_resolver
 
-WAITER_SLEEP = 180
-MAX_WAITER_ATTEMPTS = 30
+WAITER_SLEEP = 300
```
PettitWesley, line 12: Increase sleep wait time so that it prints describe task output less often (I want the full output for debugging)

DrewZhang13 (approved):

PR #690, opened 2023-06-16
integ: re-building validator images for every test run is inefficient

PettitWesley:
### Description of changes 

Currently our integ tests are failing due to an issue installing python boto3 for the integ cloudwatch validator. I'm not sure why this is happening, the error doesn't really make any sense why it'd happen repeatedly. 

```
RuntimeError: can't start new thread
Service 'cloudwatch_validator' failed to build: The command '/bin/sh -c pip3 install boto3' returned a non-zero code: 2
Test Failed for Cloudwatch (Golang).
Makefile:196: recipe for target 'integ' failed
make: *** [integ] Error 1
```

But while investigating it I realized that its really innefficient and unnecessary to re-build those images for every single test run. We might as well have them pre-built in our pipeline account. Also, when I build them locally, that works. So this should fix the current pipeline failure. 

### DIFF
--- a/buildspec_integ.yml
+++ b/buildspec_integ.yml
@@ -10,6 +10,8 @@ phases:
     commands:
       # Enforce STS regional endpoints
       - export AWS_STS_REGIONAL_ENDPOINTS=regional
+      - export CW_INTEG_VALIDATOR_IMAGE=${AWS_ACCOUNT}.dkr.ecr.us-west-2.amazonaws.com/cw-integ-validator:latest
+      - export S3_INTEG_VALIDATOR_IMAGE=${AWS_ACCOUNT}.dkr.ecr.us-west-2.amazonaws.com/s3-integ-validator:latest
       # Get the default credentials and set as environment variables
       - 'CREDS=`curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`'
       - 'export AWS_ACCESS_KEY_ID=`echo $CREDS | jq -r .AccessKeyId`'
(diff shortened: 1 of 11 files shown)


buildspec_integ.yml
```diff
@@ -10,6 +10,8 @@ phases:
     commands:
       # Enforce STS regional endpoints
       - export AWS_STS_REGIONAL_ENDPOINTS=regional
+      - export CW_INTEG_VALIDATOR_IMAGE=${AWS_ACCOUNT}.dkr.ecr.us-west-2.amazonaws.com/cw-integ-validator:latest
+      - export S3_INTEG_VALIDATOR_IMAGE=${AWS_ACCOUNT}.dkr.ecr.us-west-2.amazonaws.com/s3-integ-validator:latest
```
PettitWesley, line 14: I've built and pushed these images to our pipeline account

integ/run-integ.ps1
```diff
@@ -75,12 +75,13 @@ $env:FLUENT_BIT_IMAGE = $FluentBitImage
 $env:TAG = -join ((65..90) + (97..122) | Get-Random -Count 10 | % {[char]$_})
 # AWS_FOR_FLUENT_BIT_CONTAINER_NAME is the name for the fluent-bit container ran for each service.
 $env:AWS_FOR_FLUENT_BIT_CONTAINER_NAME = "aws-for-fluent-bit-$($env:TAG)"
+$env:CW_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/cw-integ-validator:latest"
+$env:S3_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/s3-integ-validator:latest"
```
PettitWesley, line 79: @rawahars In the linux version of this change, I parametrize the account ID, which will make it easier to run in test accounts. I use ${AWS_ACCOUNT} which I think is set by code build- this should be set in windows too right? I couldn't find any existing usages of it so I didn't add it to be safe.

DrewZhang13 (approved):

PR #692, opened 2023-06-16
load-test: windows images must differentiate server version, 2019 vs …

PettitWesley:
…2022

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/integ/run-integ.ps1
+++ b/integ/run-integ.ps1
@@ -75,8 +75,8 @@ $env:FLUENT_BIT_IMAGE = $FluentBitImage
 $env:TAG = -join ((65..90) + (97..122) | Get-Random -Count 10 | % {[char]$_})
 # AWS_FOR_FLUENT_BIT_CONTAINER_NAME is the name for the fluent-bit container ran for each service.
 $env:AWS_FOR_FLUENT_BIT_CONTAINER_NAME = "aws-for-fluent-bit-$($env:TAG)"
-$env:CW_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/cw-integ-validator-windows"
-$env:S3_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/s3-integ-validator-windows"
+$env:CW_INTEG_VALIDATOR_IMAGE = "cw-integ-validator-windows:latest"
+$env:S3_INTEG_VALIDATOR_IMAGE = "s3-integ-validator-windows:latest"
 
 # Windows specific settings.
 $env:ARCHITECTURE= "x86-64"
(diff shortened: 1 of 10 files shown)


integ/run-integ.ps1
```diff
@@ -75,8 +75,8 @@ $env:FLUENT_BIT_IMAGE = $FluentBitImage
 $env:TAG = -join ((65..90) + (97..122) | Get-Random -Count 10 | % {[char]$_})
 # AWS_FOR_FLUENT_BIT_CONTAINER_NAME is the name for the fluent-bit container ran for each service.
 $env:AWS_FOR_FLUENT_BIT_CONTAINER_NAME = "aws-for-fluent-bit-$($env:TAG)"
-$env:CW_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/cw-integ-validator-windows"
-$env:S3_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/s3-integ-validator-windows"
+$env:CW_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/cw-integ-validator-windows:$Platform-$env:ARCHITECTURE"
+$env:S3_INTEG_VALIDATOR_IMAGE = "906394416424.dkr.ecr.us-west-2.amazonaws.com/s3-integ-validator-windows:$Platform-$env:ARCHITECTURE"
```
PettitWesley: @rawahars would there be a way to support a single Windows image for both 2019 and 2022 server versions?

rawahars: Yeah we can merge the digests of both images into a single manifest. Let's discuss this offline.

DrewZhang13 (approved):

integ/run-integ.ps1
PettitWesley: From discussion, for windows images, it's better to build them each time. The base image updates every month for windows. The latest one will be cached on instance. The base image for windows is the largest part of the image. So re-build pulls in less data over network.

integ/run-integ.ps1
```diff
@@ -114,6 +114,10 @@ Function Install-Package {
         Invoke-WebRequest "https://github.com/docker/compose/releases/download/v${DockerComposeVersion}/docker-compose-Windows-x86_64.exe" -UseBasicParsing -OutFile $Env:ProgramFiles\Docker\docker-compose.exe
     }
     Write-Host $( docker-compose version )
+
+    # build validator images
+    docker build -t $env:CW_INTEG_VALIDATOR_IMAGE --file "${IntegTestRoot}/validate_cloudwatch/Dockerfile" "${IntegTestRoot}/validate_cloudwatch/"
```
rawahars, line 119: After each `docker build` command, we should check if the command succeeded or not. A check as below should do it-

```
if ($LASTEXITCODE)
{
    throw ("Failed to build $env:CW_INTEG_VALIDATOR_IMAGE image")
}
```

rawahars (changes requested):

integ/run-integ.ps1
PettitWesley: added

rawahars (approved):

PR #713, opened 2023-08-01
build: change dockerfile to set user to non-root

matthewfala:
*Issue #, if available:*

Special consideration should be taken to see if "Exec" input will still work for all common use cases. Otherwise this is a breaking change.

*Description of changes:*
Add a new user called user and switched to that user at the end of the dockerfile. This should cause fluent-bit to no-longer run as root, reducing security risks.

Test before change
```
docker run -it aws-for-fluent-bit:.../bin/bash
bash-4.2# whoami
root
```

Test after change
```
$ docker run -it amazon/aws-for-fluent-bit:main-release /bin/bash
bash-4.2$ whoami
user
```

### DIFF
--- a/scripts/dockerfiles/Dockerfile.init-release
+++ b/scripts/dockerfiles/Dockerfile.init-release
@@ -9,5 +9,8 @@ COPY --from=init-builder /fluent_bit_init_process /init/fluent_bit_init_process
 COPY init/fluent_bit_init_entrypoint.sh /init/fluent_bit_init_entrypoint.sh
 RUN chmod +x /init/fluent_bit_init_entrypoint.sh
 
+# Change to non-root user
+RUN groupadd -r user && useradd -r -g user user
+
 # Only last CMD command will be executed, automatically replaces the original entrypoint
 CMD /init/fluent_bit_init_entrypoint.sh
--- a/scripts/dockerfiles/Dockerfile.main-release
+++ b/scripts/dockerfiles/Dockerfile.main-release
@@ -54,5 +54,8 @@ RUN chmod +x /entrypoint.sh
 # Optional Metrics endpoint
 EXPOSE 2020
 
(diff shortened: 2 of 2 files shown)


scripts/dockerfiles/Dockerfile.init-release
```diff
@@ -9,5 +9,9 @@ COPY --from=init-builder /fluent_bit_init_process /init/fluent_bit_init_process
 COPY init/fluent_bit_init_entrypoint.sh /init/fluent_bit_init_entrypoint.sh
 RUN chmod +x /init/fluent_bit_init_entrypoint.sh
 
+# Change to non-root user
+RUN groupadd -r user && useradd -r -g user user
+USER user
+
```
PettitWesley: We can't do this by default. In FireLens, we need to run FLB as root. I've been meaning to create an FAQ on this for some time. I'll send you an internal note for it.

PettitWesley: Can we create the user and then optionally customers can run FLB as it at runtime if they want?

PettitWesley: I know k8s has a RunAsUser field. I'm not experienced with it though.

scripts/dockerfiles/Dockerfile.init-release
```diff
@@ -9,5 +9,8 @@ COPY --from=init-builder /fluent_bit_init_process /init/fluent_bit_init_process
 COPY init/fluent_bit_init_entrypoint.sh /init/fluent_bit_init_entrypoint.sh
 RUN chmod +x /init/fluent_bit_init_entrypoint.sh
 
+# Change to non-root user
+RUN groupadd -r user && useradd -r -g user user
+
```
PettitWesley, line 14: We need to use the `USER` keyword right?

PR #715, opened 2023-08-11
publish: suggested latest tag fix for cn partition

matthewfala:
*Issue #, if available:*
suggested latest tag fix for cn partition

Todo: test

*Description of changes:*

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -32,9 +32,10 @@ echo "Publish Latest? ${PUBLISH_LATEST}"
 
 # this code currenly works because DockerHub returns the only last 100 tags and as of March 2023 we only have 64
 # and it should keep working because dockerhub returns the latest tags first
-docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
-tag_array=(`echo ${docker_hub_image_tags}`)
-AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py linux latest ${tag_array[@]})
+public_ecr_image_tags_token=$(curl -s -S -k https://public.ecr.aws/token/ | jq -r '.token')
+public_ecr_image_tags=$(curl -s -S -k -H "Authorization: Bearer $public_ecr_image_tags_token" 'https://public.ecr.aws/v2/aws-observability/aws-for-fluent-bit/tags/list' | jq -r '.tags[]' | sort -rV)
+tag_array=(`echo ${public_ecr_image_tags}`)
+AWS_FOR_FLUENT_BIT_VERSION_PUBLIC_ECR=$(./get_latest_dockerhub_version.py linux latest ${tag_array[@]})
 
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -32,8 +32,9 @@ echo "Publish Latest? ${PUBLISH_LATEST}"
 
 # this code currenly works because DockerHub returns the only last 100 tags and as of March 2023 we only have 64
 # and it should keep working because dockerhub returns the latest tags first
-docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
-tag_array=(`echo ${docker_hub_image_tags}`)
+public_ecr_image_tags_token=$(curl -k https://public.ecr.aws/token/ | jq -r '.token')
+public_ecr_image_tags=$(curl -s -S  -k -H "Authorization: Bearer $public_ecr_image_tags_token"  'https://public.ecr.aws/v2/aws-observability/aws-for-fluent-bit/tags/list?page=1&page_size=250'  | jq -r '.tags[]' | sort -rV)
```
PettitWesley: Should make it `curl -s`? for silent

PettitWesley: How many tags do we have currently? Is hte page size a problem? I think i checked this before for docker hub and last winter we had something above 100...

PettitWesley: We checked and public ECR API call returns all tags (more than 250). Also, we only launched public ECR in Jan 2021, so there aren't all of our images in it. So at least for now, this works.

PettitWesley (approved):

scripts/publish.sh
```diff
@@ -32,17 +32,18 @@ echo "Publish Latest? ${PUBLISH_LATEST}"
 
 # this code currenly works because DockerHub returns the only last 100 tags and as of March 2023 we only have 64
 # and it should keep working because dockerhub returns the latest tags first
-docker_hub_image_tags=$(curl -s -S 'https://registry.hub.docker.com/v2/repositories/amazon/aws-for-fluent-bit/tags/?page=1&page_size=250' | jq -r '.results[].name')
-tag_array=(`echo ${docker_hub_image_tags}`)
-AWS_FOR_FLUENT_BIT_VERSION_DOCKERHUB=$(./get_latest_dockerhub_version.py linux latest ${tag_array[@]})
+public_ecr_image_tags_token=$(curl -s -S -k https://public.ecr.aws/token/ | jq -r '.token')
+public_ecr_image_tags=$(curl -s -S -k -H "Authorization: Bearer $public_ecr_image_tags_token" 'https://public.ecr.aws/v2/aws-observability/aws-for-fluent-bit/tags/list' | jq -r '.tags[]' | sort -rV)
+tag_array=(`echo ${public_ecr_image_tags}`)
+AWS_FOR_FLUENT_BIT_VERSION_PUBLIC_ECR=$(./get_latest_dockerhub_version.py linux latest ${tag_array[@]})
```
matthewfala, line 38: This is really the core of the fix

matthewfala:
Tested in all partitions and the PR resolves the cn problems and works in com and gov

PettitWesley (approved):

PR #717, opened 2023-08-14
sync tasks: add tel_aviv & melbourne account ids

matthewfala:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -124,6 +124,14 @@ hyderabad_region="ap-south-2"
 
 hyderabad_account_id="378905956269"
 
+tel_aviv_region="il-central-1"
+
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -124,6 +124,14 @@ hyderabad_region="ap-south-2"
 
 hyderabad_account_id="378905956269"
 
+tel_aviv_region="il-central-1"
```
PettitWesley, line 127: I wonder why its `il-central` ? No other `il` regions. And how did they pick `il`. 🤔

PettitWesley (approved):

scripts/publish.sh
matthewfala: israel!

PR #719, opened 2023-08-15
pipeline: verify resolve variable overwrite

matthewfala:
Sha1 is set to the sync repo's private ecr image of current version number here: https://github.com/aws/aws-for-fluent-bit/blob/120e5df2a24709a01d5138dc101636516e0dddde/scripts/publish.sh#L555 

And is compared with sha2 which is the latest image here: https://github.com/aws/aws-for-fluent-bit/blob/120e5df2a24709a01d5138dc101636516e0dddde/scripts/publish.sh#L584

These should be the same, and are compared with a verify_sha. However between setting sha1 and sha2, `verify_sha $sha1_init $sha2_init` is called which (as an error) overwrites sha1 variable to the sha1_init variable causing the future comparison of sha1 and sha2 to fail.

https://github.com/aws/aws-for-fluent-bit/blob/120e5df2a24709a01d5138dc101636516e0dddde/scripts/publish.sh#L569

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -686,10 +686,10 @@ verify_public_ecr() {
 }
 
 verify_sha() {
-	sha1=${1}
-	sha2=${2}
+	_sha1=${1}
+	_sha2=${2}
 
-	match_two_sha $sha1 $sha2
+	match_two_sha $_sha1 $_sha2
 
 	if [ "$IMAGE_SHA_MATCHED" = "TRUE" ]; then
 		echo '[Publish Verification] Successfull'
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -686,10 +686,10 @@ verify_public_ecr() {
 }
 
 verify_sha() {
-	sha1=${1}
-	sha2=${2}
+	_sha1=${1}
```
weijianz-opc, line 689: For code/doc consistency, should we name it as sha_1 ??

matthewfala: I think that would be more confusing.

So if we name sha1, it has a variable name conflict with the caller script so we can't be consistent
We can't use `sha1`

So `sha_1` and `_sha1` are currently not used in the codebase.
I prefer `_sha1` to have the underscore prefix indicate that this variable is temporary.

weijianz-opc: got it , thanks for explanation!

weijianz-opc (approved):
LGTM as just renaming variable.

PR #757, opened 2023-10-31
publish: add scanner account to sync task script

matthewfala:
Let's put ECR scanners on non prod accounts so we don't run into non-pairity issues, or break the pipeline again.

*Description of changes:*

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -136,6 +136,10 @@ gamma_region="us-west-2"
 
 gamma_account_id="626332813196"
 
+scanner_region="us-west-2"
+
+scanner_account_id="210752607241"
+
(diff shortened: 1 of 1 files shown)


PettitWesley (approved):
Why use a new account instead of just enabling scanning in main pipeline account in a different region than the pipeline (not us-west-2)?

matthewfala:
I think for something non-essential for production such as a CVE scan and chat bot, it's a good idea to isolate it to a non-production account, thus we minimize risk of breaking the production account. The production accounts should be set up with CDK. CVE scanner is setup manually currently. If we want to add this to our CDK in the future, we should add it to our production account.

matthewfala:
Need to revert. The code is making our prod account deploy to the scanner ecr, which is failing.

PR #775, opened 2023-12-06
debugging_guide: add broken connection issues guidance

matthewfala:
Add a readme section for the common keepalive timeout issue which results in broken connection errors in some customer usecases.

### DIFF
--- a/troubleshooting/debugging.md
+++ b/troubleshooting/debugging.md
@@ -5,6 +5,7 @@
 - [Understanding Error Messages](#understanding-error-messages)
     - [How do I tell if Fluent Bit is losing logs?](#how-do-i-tell-if-fluent-bit-is-losing-logs)
     - [Common Network Errors](#common-network-errors)
+    - [CloudWatch Broken Connection Error](#cloudwatch-broken-connection-error)
     - [Tail Input Skipping File](#tail-input-skipping-file)
     - [Tail Permission Errors](#tail-permission-errors)
     - [Overlimit warnings](#overlimit-warnings)
@@ -169,6 +170,24 @@ The following are common network connection errors that we have seen frequently:
 [error] [tls] SSL error: NET - Connection was reset by peer
 ```
 
+#### CloudWatch Broken Connection Error
+
+As mentioned in the section, [Common Network Errors](#common-network-errors), seeing transinent network errors such as `broken connection` are common and normal and do not indicate log loss. See [Case Studies: Rate of Network Errors](#rate-of-network-errors) for guidance on typical network error rate frequencies.
+
+```
+[error] [http_client] broken connection to <your output destination>
+```
+
+However, if you are seeing a large number of `broken_connection` errors to an AWS `cloudwatch_logs` output plugin, several per second, or around one per chunk flush, the following configuration option may resolve this issue:
+```
+[OUTPUT]
+    name cloudwatch_logs
+    ...
+    # Add the following line
+    net.keepalive_idle_timeout   4
+```
+
+Sometimes Fluent Bit fails to detect keepalive connection closures from CloudWatch Logs API resulting in the reuse of closed connections, which display `broken connection` logs. Adding the `net.keepalive_idle_timeout` config option may resolve this problem. The idle timeout value is set to `4` since CloudWatch Logs API has a `6` second keepalive idle timeout set. Fluent Bit manages open network connections every `1.5` seconds, and thus, to be safe this value is set to `4` so that Fluent Bit's network management will close keepalive connections before CloudWatch API. Note that this solution only applies to the `cloudwatch_logs` c output plugin and not the legacy `cloudwatch` go output plugin.
 
 #### Tail Input Skipping File
 


troubleshooting/debugging.md
```diff
@@ -32,6 +32,7 @@
     - [Searching old issues](#searching-old-issues)
     - [Downgrading or upgrading your version](#downgrading-or-upgrading-your-version)
     - [Network Connection Issues](#network-connection-issues)
+    - [CloudWatch Broken Connection Issues](#cloudwatch-broken-connection-issues)
```
lubingfeng: Should we move this before line 25, given this is a common issue? Do we need to update ticket template so that people can find this solution before cutting the ticket to the team?

troubleshooting/debugging.md
```diff
@@ -730,6 +731,25 @@ Remember that we also have older Golang output plugins for AWS (no longer recomm
...
 
+#### CloudWatch Broken Connection Issues
+
+As mentioned in the section, [Common Network Errors](#common-network-errors), seeing transinent network errors such as `broken connection` are common and normal and do not indicate log loss.
+
+```
+[error] [http_client] broken connection to <your output destination>
+```
+
+However, if you are seeing a large number of `broken_connection` errors to an AWS `cloudwatch_logs` output plugin, several per second, or around one per chunk flush, the following configuration option may resolve this issue:
```
PettitWesley: let's link to the case studies section to show a typical error rate to clarify large

troubleshooting/debugging.md
```diff
@@ -730,6 +731,25 @@ Remember that we also have older Golang output plugins for AWS (no longer recomm
...
+```
+[OUTPUT]
+    name cloudwatch_logs
+    ...
+    # Add the following line
+    net.keepalive_idle_timeout   4
```
PettitWesley: Let's explain the logic behind this- CW has 6 second and FLB has 1.5 second conn timer

troubleshooting/debugging.md
```diff
@@ -730,6 +731,25 @@ Remember that we also have older Golang output plugins for AWS (no longer recomm
...
+```
+
+In some situations, Fluent Bit fails to detect keepalive connection closures from Cloudwatch Logs service. Adding the `net.keepalive_idle_timeout` config option allows for Fluent Bit to shut keepalive conncetions before cloudwatch, resolving this problem.
```
PettitWesley: conncetions

troubleshooting/debugging.md
matthewfala: I think that this is the natural place for this. We may value currently over the other issues because some customers are reaching out to ask about it.

However, it may be that they are reaching out and asking about it and not the other items which are listed before it, because these items are already in our debugging guide (and more important)

troubleshooting/debugging.md
matthewfala: I'm updating this section and adding a link.

troubleshooting/debugging.md
matthewfala: Okay, please see the revision when I push it.

troubleshooting/debugging.md
matthewfala: Good catch. Thank you.

troubleshooting/debugging.md
```diff
@@ -5,6 +5,7 @@
 - [Understanding Error Messages](#understanding-error-messages)
     - [How do I tell if Fluent Bit is losing logs?](#how-do-i-tell-if-fluent-bit-is-losing-logs)
     - [Common Network Errors](#common-network-errors)
+    - [CloudWatch Broken Connection Error](#cloudwatch-broken-connection-error)
```
matthewfala, line 8: Moved the added section to higher in this debugging guide

troubleshooting/debugging.md
```diff
@@ -169,6 +170,24 @@ The following are common network connection errors that we have seen frequently:
 [error] [tls] SSL error: NET - Connection was reset by peer
 ```
 
+#### CloudWatch Broken Connection Error
+
+As mentioned in the section, [Common Network Errors](#common-network-errors), seeing transinent network errors such as `broken connection` are common and normal and do not indicate log loss. See [Case Studies: Rate of Network Errors](#rate-of-network-errors) for guidance on acceptable network error rate frequencies.
```
PettitWesley: nit: I'd say common or typical, not acceptable

troubleshooting/debugging.md
```diff
@@ -169,6 +170,24 @@ The following are common network connection errors that we have seen frequently:
...
+However, if you are seeing a large number of `broken_connection` errors to an AWS `cloudwatch_logs` output plugin, several per second, or around one per chunk flush, the following configuration option may resolve this issue:
+```
+[OUTPUT]
+    name cloudwatch_logs
+    ...
+    # Add the following line
+    net.keepalive_idle_timeout   4
+```
+
+With some configurations, Fluent Bit fails to detect keepalive connection closures from CloudWatch Logs API resulting in the reuse of closed connections, which display `broken connection` logs. Adding the `net.keepalive_idle_timeout` config option may resolve this problem. The idle timeout value is set to `4` since CloudWatch Logs API has a `6` second keepalive idle timeout set. Fluent Bit manages open network connections every `1.5` seconds, and thus, to be safe this value is set to `4` so that Fluent Bit's network management will close keepalive connections before CloudWatch API. Note that this solution only applies to the `cloudwatch_logs` c output plugin and not the legacy `cloudwatch` go output plugin.
```
PettitWesley: > With some configurations,

This leads to quesitons, if we don't know more details, then just say "Sometimes FLB fails..."

troubleshooting/debugging.md
matthewfala: Updated it to a section closer to the top of the debugging guide.

troubleshooting/debugging.md
matthewfala: Changed

troubleshooting/debugging.md
matthewfala: Updated.

lubingfeng (approved):
Good for me.

PR #795, opened 2024-03-11
Makefile target to automate linux only rebuilds

PettitWesley:
This just takes the manual step we currently perform to change the version and adds a script which will change the version instead.

### DIFF
--- a/scripts/linux-rebuild.sh
+++ b/scripts/linux-rebuild.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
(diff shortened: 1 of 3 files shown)


scripts/linux-rebuild.sh
```diff
@@ -0,0 +1,33 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+set -xeuo pipefail
+
+scripts=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+cd "${scripts}"
+
+# If latest is a rebuild ("2.31.12.20230911") returns the base ("2.31.12")
+export AWS_FOR_FLUENT_BIT_BASE_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION | cut -d '.' -f 1-3)
```
PettitWesley, line 21: I need to parameterize this so that it can rebuild either latest or stable in the pipeline

PettitWesley: I will include this in my pipeline PR

Issue #746, opened 2023-10-17
CVE-2023-38545 and CVE-2023-38545 vulnerabilities


egorchabala:
Multiple security scanning tools reported that `aws-for-fluentbit` docker image might be vulnerable to the following vulnerability:

-  [CVE-2023-38545](https://alas.aws.amazon.com/cve/html/CVE-2023-38545.html)
-  [CVE-2023-38546](https://alas.aws.amazon.com/cve/html/CVE-2023-38546.html)

Is there a schedule for a new release with a patched base image?

PettitWesley:
I got this:

```
147dda285fd4:fluent-bit wppttt$ trivy image public.ecr.aws/aws-observability/aws-for-fluent-bit:2.31.12.20231011
2023-10-18T17:56:29.579-0700	INFO	Vulnerability scanning is enabled
2023-10-18T17:56:29.579-0700	INFO	Secret scanning is enabled
2023-10-18T17:56:29.579-0700	INFO	If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2023-10-18T17:56:29.579-0700	INFO	Please see also https://aquasecurity.github.io/trivy/v0.38/docs/secret/scanning/#recommendation for faster secret detection
2023-10-18T17:57:49.052-0700	INFO	Detected OS: amazon
2023-10-18T17:57:49.052-0700	INFO	Detecting Amazon Linux vulnerabilities...
2023-10-18T17:57:49.065-0700	INFO	Number of language-specific files: 0

public.ecr.aws/aws-observability/aws-for-fluent-bit:2.31.12.20231011 (amazon 2 (Karoo))

Total: 1 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 1, CRITICAL: 0)

┌────────────┬────────────────┬──────────┬────────────────────┬────────────────────┬──────────────────────────────────────────────────────────────┐
│  Library   │ Vulnerability  │ Severity │ Installed Version  │   Fixed Version    │                            Title                             │
├────────────┼────────────────┼──────────┼────────────────────┼────────────────────┼──────────────────────────────────────────────────────────────┤
│ libnghttp2 │ CVE-2023-44487 │ HIGH     │ 1.41.0-1.amzn2.0.3 │ 1.41.0-1.amzn2.0.4 │ Multiple HTTP/2 enabled web servers are vulnerable to a DDoS │
│            │                │          │                    │                    │ attack (Rapid...                                             │
[...]

PR #860, opened 2024-10-07
Revert "lock to go 1.20.7"

sparrc:
This reverts commit 7ea28b0a41d56fae96aeee32d821f2b00402d467.

*Issue #, if available:*

https://github.com/aws/aws-for-fluent-bit/issues/746

*Description of changes:*

This commit was originally added to pin the golang build version to 1.20.7 because of an issue in golang related to CGO: https://github.com/golang/go/issues/62130#issuecomment-1684431260

This issue was fixed for most systems, although the issue is still open because it appears to have opened up a separate issue with MUSL-based systems (such as alpine): https://github.com/golang/go/issues/68285

Since the fluent-bit container is on a glibc-based system (Amazon Linux), we should be safe to revert back to building with the latest version of golang.

### DIFF
--- a/scripts/build_plugins.sh
+++ b/scripts/build_plugins.sh
@@ -166,13 +166,10 @@ fi
 # get Go stable version
 # Dockerfiles do not allow env vars to be set by commands
 # and persist from one command to the next
-# GO_OUTPUT=$(curl --silent https://go.dev/VERSION?m=text | cut -d "o" -f 2)
-# OLD_IFS=$IFS
-# IFS=$'\n' GO_STABLE_VERSION=($GO_OUTPUT)
-# IFS=$OLD_IFS
-# 2023-08-21: pinning to 1.20.7 go as 1.21.0 fails with FLB
-# https://github.com/golang/go/issues/62130#issuecomment-1684431260
-export GO_STABLE_VERSION="1.20.7"
+GO_OUTPUT=$(curl --silent https://go.dev/VERSION?m=text | cut -d "o" -f 2)
+OLD_IFS=$IFS
+IFS=$'\n' GO_STABLE_VERSION=($GO_OUTPUT)
+IFS=$OLD_IFS
 echo "Using go:stable version ${GO_STABLE_VERSION}"
 PLUGIN_BUILD_ARGS="$PLUGIN_BUILD_ARGS --build-arg GO_STABLE_VERSION=${GO_STABLE_VERSION}"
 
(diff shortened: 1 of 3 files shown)


swapneils (approved):
Looks good! Thanks for also checking the scope of the Golang-side changes.

We can publish this in the upcoming `2.32.3`.

aaithal:
thanks! i do have one question though. I see that we're using AL2 as the base image (Switching to AL2023 is topic for another day). Why would we not want to install the golang version that's packaged in AL2 repos vs doing our own "stable" determinations? What's the harm in doing something like `RUN yum install -y golang`? The advantage is that we don't have to figure out a CVE/patching story for golang ourselves. We have AL maintainers for that.

sparrc:
> thanks! i do have one question though. I see that we're using AL2 as the base image (Switching to AL2023 is topic for another day). Why would we not want to install the golang version that's packaged in AL2 repos vs doing our own "stable" determinations? What's the harm in doing something like `RUN yum install -y golang`? The advantage is that we don't have to figure out a CVE/patching story for golang ourselves. We have AL maintainers for that.

I think there are two main advantages to building with the latest golang version for now:

1. We get onto the latest version of golang with all CVE fixes as soon as possible and avoid having to wait on Amazon Linux releasing the latest CVE patches to their yum repos.
2. We are made aware as soon as possible when the latest golang version might have some change, bug, or incompatibility with our build process.

So far we have been able to rely on the latest golang version being stable and not causing us any major issues so I don't think we currently have any strong reason to hold back on picking up the latest as soon as it's available.

aaithal:
> > thanks! i do have one question though. I see that we're using AL2 as the base image (Switching to AL2023 is topic for another day). Why would we not want to install the golang version that's packaged in AL2 repos vs doing our own "stable" determinations? What's the harm in doing something like `RUN yum install -y golang`? The advantage is that we don't have to figure out a CVE/patching story for golang ourselves. We have AL maintainers for that.
> 
> I think there are two main advantages to building with the latest golang version for now:
> 
> 1. We get onto the latest version of golang with all CVE fixes as soon as possible and avoid having to wait on Amazon Linux releasing the latest CVE patches to their yum repos.
> 2. We are made aware as soon as possible when the latest golang version might have some change, bug, or incompatibility with our build process.
> 
> So far we have been able to rely on the latest golang version being stable and not causing us any major issues so I don't think we currently have any strong reason to hold back on picking up the latest as soon as it's available.

I don't want to hold off this PR for now. But, I do think we need to revisit this. Supply chain integrity is an important aspect of security and by not relying on AL vended golang binaries, we are vulnerable to mitm attacks or attackers compromising 3rd party repos that we're depending on for our golang binaries.

If we're out of compliance because AL hasn't caught up with CVEs, we have bigger issues at that point anyway because the container image as a whole would not be considered compliant. 

But as I said earlier, I don't want to hold off this PR for now. But, we need to revisit how we consume all of our dependencies more holistically.

aaithal (approved):

PR #874, opened 2024-11-12
Put BASIC registry scanning config during release

sparrc:
### Summary

Adds `aws ecr put-registry-scanning-configuration --scan-type BASIC --region ${region}` command to put registry scanning config.

This is necessary because we run a one-time scan during the release, and this feature is only available if BASIC scanning is set on the registry as a whole.

Also ran a shell formatter script (shfmt) to get rid of all the random whitespace in this file.

*Issue #, if available:* NA

### Testing

This command was run during release 2.32.4. Confirmed that subsequent duplicate calls to this command also succeed.

NA

### Licensing


sparrc:
closing in favor of https://github.com/aws/aws-for-fluent-bit/pull/897

PR #897, opened 2025-02-12
scripts: generate-changelog and put BASIC registry scanning

sparrc:
New script added to generate release notes for aws-for-fluent-bit releases.

Also adds `aws ecr put-registry-scanning-configuration --scan-type BASIC --region ${region}` command to put registry scanning config.

This is necessary because we run a one-time scan during the release, and this feature is only available if BASIC scanning is set on the registry as a whole.

### Testing

generate-changelog script was run during the fluent-bit 2.32.4.20241217 and 2.32.5.20250212 releases

scanning config command was run during release 2.32.4. Confirmed that subsequent duplicate calls to this command also succeed.

NA

### Licensing

### DIFF
--- a/scripts/generate_changelog.sh
+++ b/scripts/generate_changelog.sh
@@ -32,7 +32,7 @@ while true; do
 done
 
 # Find the most recent AL2 tag
-most_recent_al2=$(printf '%s\n' "${all_tags[@]}" | grep '^2\.' | grep -v arm | grep -v amd | sort -V | tail -n 1)
+most_recent_al2=$(printf '%s\n' "${all_tags[@]}" | grep '^2\.' | grep -v minimal | grep -v arm | grep -v amd | sort -V | tail -n 1)
 
 # Read the JSON file
 json_file="linux.version"
@@ -55,7 +55,7 @@ This release includes:
 * Amazon Kinesis Firehose for Fluent Bit ${firehose_plugin_version#v}
 * Amazon Linux base container image version: $most_recent_al2
 
-Changes introduced the fluent-bit container:
+Compared to the previous release, this release adds:
 * Fix - TODO blah blah [#TODO](https://github.com/amazon-contributing/upstream-to-fluent-bit/pull/TODO)
 * Enhancement - TODO blah blah [#TODO](https://github.com/aws/aws-for-fluent-bit/pull/TODO)
 EOF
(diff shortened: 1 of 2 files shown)


scripts/generate_changelog.sh
```diff
@@ -0,0 +1,61 @@
...
+cat << EOF
+### $version
+This release includes:
+* Fluent Bit [$fluent_bit_version](https://github.com/fluent/fluent-bit/tree/v$fluent_bit_version)
+* Amazon CloudWatch Logs for Fluent Bit ${cloudwatch_plugin_version#v}
+* Amazon Kinesis Streams for Fluent Bit ${kinesis_plugin_version#v}
+* Amazon Kinesis Firehose for Fluent Bit ${firehose_plugin_version#v}
+* Amazon Linux base container image version: $most_recent_al2
+
+Changes introduced the fluent-bit container:
```
swapneils: Can we change this line to "Compared to <previous version> this release adds:"? Or query the first line matching "^### " in CHANGELOG.md to fill in the actual <previous version> value? 

Prefer to keep this consistent with prior releases unless there's reason to change it.

scripts/generate_changelog.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    # Check if there's a next token
+    next_token=$(echo "$response" | jq -r '.nextToken')
+    if [[ "$next_token" == "null" ]]; then
+        break
+    fi
+done
+
+# Find the most recent AL2 tag
+most_recent_al2=$(printf '%s\n' "${all_tags[@]}" | grep '^2\.' | grep -v arm | grep -v amd | sort -V | tail -n 1)
```
swapneils: Nit: Can we add `grep -v minimal` in case AL2 starts using `minimal` images like AL2023?

swapneils (changes requested):

scripts/generate_changelog.sh
sparrc: done

scripts/generate_changelog.sh
sparrc: done

scripts/generate_changelog.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
```
singholt, line 3: add `set -xeuo pipefail`

singholt (approved):

scripts/generate_changelog.sh
sparrc: done

singholt (approved):

swapneils (approved):

PR #912, opened 2025-03-19
Partial CVE automation

swapneils:
### Summary

This script restructures some scripts, e.g. `generate_changelog.sh`, and adds a python3 script `publish_cve_update.py` which automates the process of checking the remote AL2 repository and (if needed) generating a commit to release a new `latest` version. The commit is made in a new local branch named `patch-automation-X` where `X` is the same date extension that would be used for the CVE update.

Note that this change verifies whether there have been any commits to `aws-for-fluent-bit` or `upstream-to-fluent-bit` since the last release (will change to fluent-bit upstream later), and if so attempts to infer change descriptions for these changes. If a CVE release has the string `Compared to the previous release, this release adds:`, **even if the section is empty**, please check both this repository and its upstream to ensure no unintended changes are being released.

If a release is not a CVE release, the operator will need to modify the version number in `linux.version`, `AWS_FOR_FLUENT_BIT_VERSION`, and `changelog.md`.

This script also pushes the new branch to the `origin` remote of your local cloned repository. Note that we validate the attempt to switch branches and said action is idempotent, so this should not ever result in pushing to origin/mainline unless some other process switches the branch partway through this script.

Usage steps (after merging this PR):
1. Go to the root directory of your clone of `aws-for-fluent-bit`
2. Switch to the `mainline` branch
3. Run `python3 ./scripts/publish_cve_update.py`
4. Monitor the script's output for shell script failures. If these occur, the local environment or the script will need to be debugged.
5. You should end up in the `mainline` branch, with a new `patch-automation-X` branch existing in your local clone of the repository as well as the remote origin.

*Issue #, if available:*

### Testing

`make debug` succeeded: N/A
Integ tests succeeded: N/A
New tests cover the changes: N/A

### Licensing

### DIFF
--- a/scripts/generate_changelog.sh
+++ b/scripts/generate_changelog.sh
@@ -1,39 +1,8 @@
 #!/bin/bash
 set -xeuo pipefail
 
-# Initialize variables
-next_token=""
-all_tags=()
-
-# Get all amazonlinux container image tags
-while true; do
-    if [ -z "$next_token" ]; then
-        response=$(curl -sSL \
-            --header "Content-Type: application/json" \
-            --request POST \
-            --data '{"registryAliasName":"amazonlinux","repositoryName":"amazonlinux","maxResults":250}' \
-            https://api.us-east-1.gallery.ecr.aws/describeImageTags)
-    else
-        response=$(curl -sSL \
-            --header "Content-Type: application/json" \
-            --request POST \
-            --data "{\"registryAliasName\":\"amazonlinux\",\"repositoryName\":\"amazonlinux\",\"nextToken\":\"$next_token\",\"maxResults\":250}" \
-            https://api.us-east-1.gallery.ecr.aws/describeImageTags)
-    fi
-
-    # Extract tags and add them to the array
-    tags=$(echo "$response" | jq -r '.imageTagDetails[].imageTag')
-    all_tags+=($tags)
-
-    # Check if there's a next token
-    next_token=$(echo "$response" | jq -r '.nextToken')
-    if [[ "$next_token" == "null" ]]; then
-        break
-    fi
-done
-
 # Find the most recent AL2 tag
-most_recent_al2=$(printf '%s\n' "${all_tags[@]}" | grep '^2\.' | grep -v minimal | grep -v arm | grep -v amd | sort -V | tail -n 1)
+most_recent_al2=$(./scripts/most_recent_al2.sh)
 
 # Read the JSON file
 json_file="linux.version"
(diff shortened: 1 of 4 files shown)


scripts/generate_changelog.sh
```diff
@@ -1,39 +1,42 @@
...
+
+#     # Check if there's a next token
+#     next_token=$(echo "$response" | jq -r '.nextToken')
+#     if [[ "$next_token" == "null" ]]; then
+#         break
+#     fi
+# done
+
+# # Find the most recent AL2 tag
+# most_recent_al2=$(printf '%s\n' "${all_tags[@]}" | grep '^2\.' | grep -v minimal | grep -v arm | grep -v amd | sort -V | tail -n 1)
```
ShelbyZ: Consider removing this commented code as it is in git history and moved to the new shell script

scripts/generate_changelog.sh
```diff
@@ -46,17 +49,107 @@ cloudwatch_plugin_version=$(echo "$json_content" | jq -r '.linux."cloudwatch-plu
 kinesis_plugin_version=$(echo "$json_content" | jq -r '.linux."kinesis-plugin"')
 firehose_plugin_version=$(echo "$json_content" | jq -r '.linux."firehose-plugin"')
 
+changelog_commit=$(git log -n1 --pretty=format:%h CHANGELOG.md)
+current_commit=$(git log -n1 --pretty=format:%h)
+changelog_commit="9c8600e"
```
ShelbyZ: Should this be removed?

scripts/generate_changelog.sh
```diff
@@ -46,17 +49,107 @@ cloudwatch_plugin_version=$(echo "$json_content" | jq -r '.linux."cloudwatch-plu
...
+
+compared_string=""
```
ShelbyZ, line 35: Do you have an example of the output this generates?

scripts/publish_cve_update.py
```diff
@@ -0,0 +1,126 @@
...
+#     upstream_repo = None
+
+# Go to a new branch for this automation.
+numeric_date_string = datetime.now().strftime("%Y%m%d")
+run_command(["/usr/bin/git", "fetch".format(numeric_date_string)],
+            fail_on_error=True, tries=2)
+run_command(["/usr/bin/git", "switch", "-C", "patch-automation-{}".format(numeric_date_string)],
+            fail_on_error=True)
+# Reset the branch to master in case it isn't already
+# run_command(["/usr/bin/git", "reset", "--hard", "origin/mainline"])
```
ShelbyZ: Should we remove this and the comment above?

scripts/publish_cve_update.py
```diff
@@ -0,0 +1,126 @@
...
+        print(f"Failed to run command {command_arr}, quitting")
+        quit()
+
+
+print("ls in default dir")
+print(run_command(["ls"]))
+print("ls in fluent bit tmp dir")
+print(run_command(["ls"], cwd="/tmp"))
+
+print("goodbye")
```
ShelbyZ: Are these still needed or just general debugging?

scripts/publish_cve_update.py
```diff
@@ -0,0 +1,126 @@
...
+#     print(run_command(["whereis", "git"]))
+#     clone_output = run_command(["git", "clone", upstream_uri, upstream_repo],
+#                                fail_on_error=False, timeout=30)
+#     if clone_output is not None:
+#         run_command(["git", "checkout", "1.9.10"], cwd=upstream_repo,
+#                     fail_on_error=True)
+#         break
+# else:
+#     print("Couldn't clone FLB upstream. Will ignore upstream commits.")
+#     upstream_repo = None
```
ShelbyZ: Let's remove comment code

ShelbyZ (changes requested):
Some initial comments, may need some additional time on the py script

scripts/publish_cve_update.py
swapneils: Removed

scripts/publish_cve_update.py
swapneils: This part was an optional enhancement (checking for new commits in the upstream when  which I'd commented for test purposes. Uncommented, working on the changes to actually check it but it's non-blocking for this PR.

scripts/generate_changelog.sh
swapneils: This only outputs a single line with the latest commit ID. For instance, running it now in my local cloned repo:
```
384c6c7
```

scripts/publish_cve_update.py
swapneils: This was commented pending better validation of the branch switch. Added validations and uncommented, so now the script should be idempotent w.r.t. effects on the created branch (though note that for now we still create a new branch if the date changes).

scripts/generate_changelog.sh
swapneils: Removed the hardcoding of `changelog_commit` and validated it still detects that new commits exist.

Note that the variables themselves can't be removed as they gate the addition of the "Compared to the previous release" line in the Changelog.

scripts/generate_changelog.sh
swapneils: Removed.

scripts/publish_cve_update.py
swapneils: Fixed, we now correctly clone the upstream and try to parse commits there into the CHANGELOG.

scripts/generate_changelog.sh
```diff
@@ -1,39 +1,11 @@
...
-    all_tags+=($tags)
-
-    # Check if there's a next token
-    next_token=$(echo "$response" | jq -r '.nextToken')
-    if [[ "$next_token" == "null" ]]; then
-        break
-    fi
-done
+# # Find the most recent AL2 tag
+# most_recent_al2=$(printf '%s\n' "${all_tags[@]}" | grep '^2\.' | grep -v minimal | grep -v arm | grep -v amd | sort -V | tail -n 1)
```
ShelbyZ: Isn't this just duplicated from check_for_new_al_version?

scripts/publish_cve_update.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python3
```
ShelbyZ, line 1: Was there a reason to favor a python script over another shell script?

sparrc:
> Note that this change verifies whether there have been any commits to aws-for-fluent-bit or upstream-to-fluent-bit since the last release (will change to fluent-bit upstream later), and if so attempts to infer change descriptions for these changes. If a CVE release has the string Compared to the previous release, this release adds:, even if the section is empty, please check both this repository and its upstream to ensure no unintended changes are being released.

For the purposes of CVE-only releases, I'm not sure if we should be doing this much logic and parsing. It seems to me that at a high level we should simply do something like:

1. Checkout most recent tagged commit (ex. `v2.32.5.20250305`).
2. Update base Amazon Linux container version.
3. Update release/tag version (ex. `v2.32.5.20250324`)
4. Run release (maybe from a branch to avoid picking up unwanted changes to mainline).

Is this something that was considered? I am wondering if we should separate the "regular" releases that infer change descriptions et al into a separate change from the CVE-only releases?

swapneils:
> Is this something that was considered? I am wondering if we should separate the "regular" releases that infer change descriptions et al into a separate change from the CVE-only releases?

For CVE releases:

The dependency on checking the commit history at all isn't one we can remove, since for CVE releases we need to know we're not accidentally releasing a feature change.

Given that we're going to parse the commits regardless, the additional logic to guess what those commits mean improves debuggability if we are accidentally releasing change commits, especially if we're running this script via automations and just double-checking the results. The investment to add these components is only writing the code, so it doesn't add any additional ops burden either.

For feature releases:

I don't think we will fully automate feature releases, at least at this stage of the pipeline. In addition to the changelog modifications, they also need to bump the correct component of our semantic versioning, and manual checks through the relevant repos are also useful as a final pre-integ-test spot-check to avoid publishing bad changes.

If we take _partial_ automation of feature releases as the goal, there seems no reason to separate them into a separate script, as the automateable parts of the process are generally shared with CVE releases. The more advanced parts of the commit parsing are the only components essential to feature releases but not CVE releases, and as mentioned above they _do_ improve the CVE release experience as well, so inclined to keep them here.

Note: Even if we _do_ make separate script endpoints for feature vs CVE releases, still inclined for this logic to stay in generate_changelog.sh and be referenced by both automations, to avoid branching logic or duplicating `generate_changelog.sh` for each release type.

scripts/generate_changelog.sh
swapneils: This is looking at an outdated code-version. We use the new al version script to populate this variable

scripts/publish_cve_update.py
swapneils: The Python script has more complex control flow (e.g. more `for` statements, retries and timeouts on CLI commands) and used `datetime` (I don't know that datetime syntax is guaranteed-same across Mac vs AL2 vs Cloud Desktop), so using Python gave better maintainability and development-speed.

ShelbyZ (approved):

PR #942, opened 2025-06-23
Fix init process s3 getBucketLocation issue

### LINKED ISSUE #929: Can't load init files from S3 in eu-west-1, wrong S3 URL
### Describe the question/issue
When running an ECS task in eu-west-1 (Ireland), using a firelens configuration that uses s3 init from a file in a bucket in eu-west-1, it attempts to pull the file from https://s3.EU.amazonaws.com/...

This is because the init Go process is using GetBucketLocation to get the region where the bucket is located, but for some reason only in eu-west-1, the LocationConstraint is "EU" instead of the region name. Consider the following awscli output:

```
❯ aws s3api --profile prod --region eu-west-1 get-bucket-location --bucket "(my bucket in eu-west-1)"
{
    "LocationConstraint": "EU"
}
 
❯ aws s3api --profile prod --region eu-west-1 get-bucket-location --bucket "(my bucket in us-west-2)"
{
    "LocationConstraint": "us-west-2"
}
```

I see in the [init process code](https://github.com/aws/aws-for-fluent-bit/blob/mainline/init/fluent_bit_init_process.go#L242-L244) it has special handling for us-east-1 (whose LocationConstraint is an empty string).  I think we need to add special handling for eu-west-1 as well.

UPDATE: I found out that on very old buckets (from around the time the eu-west-1 region was first launched), LocationConstraint is EU, but on newer buckets, it's eu-west-1.

UPDATE 2: This odd behavior is [mentioned in the GetBucketLocation docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html):

Image

ShelbyZ:
### Summary
For eu-west-1 region s3:GetBucketLocation returns EU for LocationConstraint which is not currently handled. This change adds a check to convert EU --> eu-west-1.

Fixes #929

> Buckets in Region us-east-1 have a LocationConstraint of null. Buckets with a LocationConstraint of EU reside in eu-west-1.

Our code previously had a check for us-east-1 and null but lacked checking for EU. Also added a todo for migrating to s3:HeadBucket (larger effort) as a follow-up.

### Testing
Local testing

`make debug` succeeded: yes
Integ tests succeeded: yes

Fix init process s3 getBucketLocation issue for eu-west-1 region

### Licensing

### DIFF
--- a/init/fluent_bit_init_process.go
+++ b/init/fluent_bit_init_process.go
@@ -225,6 +225,9 @@ func getS3ConfigFile(userInput string) string {
 	bucketName := bucketAndFile[0]
 	s3FilePath := bucketAndFile[1]
 
+	// TODO: migrate to s3:HeadBucket and use BucketRegion
+	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html
+	// https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Client.HeadBucket
 	// get bucket region
 	input := &s3.GetBucketLocationInput{
 		Bucket: aws.String(bucketName),
@@ -238,9 +241,13 @@ func getS3ConfigFile(userInput string) string {
 
 	bucketRegion := string(output.LocationConstraint)
 	// Buckets in Region us-east-1 have a LocationConstraint of null
-	// https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#GetBucketLocationOutput
-	if bucketRegion == "" {
+	// Buckets in Region eu-west-1 have a LocationConstraint of EU
+	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html#API_GetBucketLocation_ResponseSyntax
+	switch bucketRegion {
+	case "":
 		bucketRegion = "us-east-1"
+	case "EU":
+		bucketRegion = "eu-west-1"
 	}
 
 	// create a downloader


singholt (approved):

init/fluent_bit_init_process.go
```diff
@@ -238,9 +241,13 @@ func getS3ConfigFile(userInput string) string {
...
 	bucketRegion := string(output.LocationConstraint)
 	// Buckets in Region us-east-1 have a LocationConstraint of null
-	// https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#GetBucketLocationOutput
-	if bucketRegion == "" {
+	// Buckets in Region eu-west-1 have a LocationConstraint of EU
+	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html#API_GetBucketLocation_ResponseSyntax
+	switch bucketRegion {
+	case "":
 		bucketRegion = "us-east-1"
+	case "EU":
```
mye956, line 249: Might be worth adding unit tests that covers "EU" and a few other regions as well. WDYT?

mye956: Happy for this to be done as a follow up but probably safer to have some form of unit testing for this if possible imo

mye956 (approved):

init/fluent_bit_init_process.go
ShelbyZ: I'll see about a follow-up PR with added testing as it has become a bit of a mountain to rework the existing method to make it more testable.

PR #943, opened 2025-06-23
Add FLB_REPOSITORY to allow overriding default Fluent Bit Repo

ShelbyZ:
### Summary
- Added FLB_REPOSITORY to Makefile with default for upstream
- Added --build-arg for Dockerfile.build
- Added FLB_REPOSITORY ARG to Dockerfile.build
- Rework git clone command to select a single branch to checkout for Fluent Bit

### Testing
Local testing/builds

`make debug` succeeded: yes
Integ tests succeeded: yes

Add FLB_REPOSITORY to allow overriding default Fluent Bit Repo

### Licensing

### DIFF
--- a/scripts/dockerfiles/Dockerfile.build
+++ b/scripts/dockerfiles/Dockerfile.build
@@ -2,6 +2,8 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder
 
 # Fluent Bit version; update these for each release
 ENV FLB_VERSION 1.9.10
+# Fluent Bit repository to checkout
+ARG FLB_REPOSITORY ${FLB_REPOSITORY}
 # branch to pull parsers from in github.com/fluent/fluent-bit-docker-image
 ENV FLB_DOCKER_BRANCH 1.8
 
@@ -76,9 +78,8 @@ FROM builder as compile
 
 # Get Fluent Bit source code
 WORKDIR /tmp/fluent-bit-$FLB_VERSION/
-RUN git clone https://github.com/amazon-contributing/upstream-to-fluent-bit.git /tmp/fluent-bit-$FLB_VERSION/
+RUN git clone --single-branch --branch $FLB_VERSION ${FLB_REPOSITORY} /tmp/fluent-bit-$FLB_VERSION/
 WORKDIR /tmp/fluent-bit-$FLB_VERSION/build/
-RUN git checkout $FLB_VERSION
 
 # Apply Fluent Bit patches to base version
 COPY AWS_FLB_CHERRY_PICKS \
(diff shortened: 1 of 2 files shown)


scripts/dockerfiles/Dockerfile.build
```diff
@@ -76,9 +78,8 @@ FROM builder as compile
 
 # Get Fluent Bit source code
 WORKDIR /tmp/fluent-bit-$FLB_VERSION/
-RUN git clone https://github.com/amazon-contributing/upstream-to-fluent-bit.git /tmp/fluent-bit-$FLB_VERSION/
+RUN git clone --single-branch --branch $FLB_VERSION ${FLB_REPOSITORY} /tmp/fluent-bit-$FLB_VERSION/
```
singholt, line 81: IIUC, the upstream fluent-bit repo uses tags instead of branch, for each version. I think we either need to switch our fork to checking out tags to be consistent with upstream, or update this command to be conditional depending on the repo.

scripts/dockerfiles/Dockerfile.build
```diff
@@ -2,6 +2,8 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder
 
 # Fluent Bit version; update these for each release
 ENV FLB_VERSION 1.9.10
+# Fluent Bit repository to checkout
+ARG FLB_REPOSITORY ${FLB_REPOSITORY}
```
singholt, line 6: Please update the README

scripts/dockerfiles/Dockerfile.build
ShelbyZ: I had to look this up but `git clone`'s `--branch` switch takes either branches or tags - https://git-scm.com/docs/git-clone

> --branch <name>
Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.

singholt: Nice. I would suggest just testing it once with upstream repo, unless you've done it already

ShelbyZ: Ha, I definitely did both tags/branches without issues on both upstream/fluent-bit repositories. For this CR I directly tested against the existing upstream/branch combo.

scripts/dockerfiles/Dockerfile.build
ShelbyZ: I have a second follow-up change after this that will introduce building select tags/branches in a similar approach. I can have that PR add a section on customized build with repository/branch or tag override into README.md

singholt (approved):

mye956 (approved):

PR #875, opened 2024-11-13
Remove 'gimme' tool and -devel packages from build toolchain

sparrc:
### Summary

Instead use official public ECR golang images.

Also remove unecessary "-devel" packages in main-release image.

Fixes https://github.com/aws/aws-for-fluent-bit/issues/368

This is a re-attempt of https://github.com/aws/aws-for-fluent-bit/pull/868, except this time only remove devel packages from the "main release base" image, leave them in the build image. And this time I've also run `make debug`. 

*Issue #, if available:* 368

### Testing

`make debug` succeeded: yes
Integ tests succeeded: yes 
New tests cover the changes: no

### Description for the changelog

Remove 'gimme' build tool, remove some -devel packages from release base image

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #966, opened 2025-07-25
Pass AL_TAG to build_plugins.sh

ShelbyZ:
### Summary
Fix issue that AL_TAG is not passed to build_plugins.sh script. This prevents overriding the AL_TAG.

Minor re-org for linux-plugins/windows-plugins export definition

### Testing
Tested by building local plugins and validating that AL_TAG override:
1. Defaults to 2 when not set
2. Allows override values provided AL2023, specific AL versions
3. Is passed to docker build command (--build-arg AL_TAG)
4. latest/init/stable images run without issue

`make debug` succeeded: yes
Integ tests succeeded: yes

Pass AL_TAG to build_plugins.sh

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,8 @@ build:
 build-init:
 	docker build $(DOCKER_BUILD_FLAGS) --build-arg AL_TAG=${AL_TAG} -t amazon/aws-for-fluent-bit:build-init -f ./scripts/dockerfiles/Dockerfile.build-init .
 
-#TODO: the bash script opts does not work on developer Macs
-windows-plugins: export OS_TYPE = windows
-linux-plugins: export OS_TYPE = linux
-
 .PHONY: windows-plugins
+windows-plugins: export OS_TYPE = windows
 windows-plugins:
 	./scripts/build_plugins.sh \
     	--KINESIS_PLUGIN_CLONE_URL=${KINESIS_PLUGIN_CLONE_URL} \
@@ -65,6 +62,7 @@ windows-plugins:
     	--DOCKER_BUILD_FLAGS=${DOCKER_BUILD_FLAGS}
 
 .PHONY: linux-plugins
+linux-plugins: export OS_TYPE = linux
 linux-plugins:
 	./scripts/build_plugins.sh \
     	--KINESIS_PLUGIN_CLONE_URL=${KINESIS_PLUGIN_CLONE_URL} \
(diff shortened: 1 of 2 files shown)


prateekchaudhry (approved):

Makefile
```diff
@@ -76,7 +74,8 @@ linux-plugins:
     	--CLOUDWATCH_PLUGIN_CLONE_URL=${CLOUDWATCH_PLUGIN_CLONE_URL} \
     	--CLOUDWATCH_PLUGIN_TAG=${CLOUDWATCH_PLUGIN_TAG} \
     	--CLOUDWATCH_PLUGIN_BRANCH=${CLOUDWATCH_PLUGIN_BRANCH} \
-    	--DOCKER_BUILD_FLAGS=${DOCKER_BUILD_FLAGS}
+    	--DOCKER_BUILD_FLAGS=${DOCKER_BUILD_FLAGS} \
+		--AL_TAG=${AL_TAG}
```
singholt, line 78: nit: indentation seems off

scripts/build_plugins.sh
```diff
@@ -107,6 +109,12 @@ do
     --DOCKER_BUILD_FLAGS)
       if [ -n "$2" ];then PLUGIN_BUILD_ARGS="$PLUGIN_BUILD_ARGS $2";fi
       shift 2;;
+    --AL_TAG)
+      if [ -n "$2" ]; then
+        PLUGIN_BUILD_ARGS="$PLUGIN_BUILD_ARGS --build-arg AL_TAG=$2"
```
singholt, line 114: Where/how is this eventually used?

Makefile
ShelbyZ: Huh, whitespace mysteries, can fix

scripts/build_plugins.sh
ShelbyZ: The -AL_TAG override will be used when calling make-release/make-debug which currently assigns a default 2 in the makefile. For AL2023, we will override that default to AL2023 and it will flow down to all make targets and get passed into this script `--AL_TAG=${AL_TAG}` == `--AL_TAG=2023`.

I tried to cut-up the larger PR for all of the changes to make it easier to focus and not get lost in details

singholt: I see. But why do the plugin builds need to be AL aware? We're doing `go build` for each of the Go plugins we bake in the image right?

ShelbyZ: I suppose that is true, they do not need to be aware. We go build each plugin and then source that into the final image

ShelbyZ:
Synced offline, and while it would be nice to have AL_TAG controlling the AL version for plugin builds. It is not required and should not effect the final runtime image.

There is already a pull request related to removing the gimme/go approach in favor of an image using specific go version - https://github.com/aws/aws-for-fluent-bit/pull/875, which is the better long term approach

PR #974, opened 2025-08-07
Add AL2023 dockerfiles

ShelbyZ:
### Summary
- Add support for AL2023 through dependency dockerfiles
- Fix core_uploader.sh check for no core files via fuzzy matching

Note: no new makefile targets are introduced using the AL2023 dockerfiles, this will be part of a follow-up commits

AL2023 build dependencies only have a few minor differences from AL2:
- cmake replaces cmake3 (version is 3.x)
- openssl-devel replaces openssl11-devel
- pkgconf-pkg-config replaces pkgconfig
- No need to update alternatives to use cmake

AL2023 runtime dependencies follow https://docs.aws.amazon.com/linux/al2023/ug/barebones-containers.html to install a small set of a dependencies that are copied into a scratch image reducing overall vulnerability surface area.

### Testing
- Tested using AL_TAG overrides for release/debug/debug-valgrind Makefile targets to generate AL2023 images
- Tested changes through pipeline build/deploy/integ test for AL2023 images
- Tested the core_uploader.sh change for both AL2 and AL2023 to ensure no backwards compatibility issues (both detected no core files)

`make debug` succeeded: yes
Integ tests succeeded: yes

Add AL2023 dockerfiles

### Licensing

### DIFF
--- a/scripts/dockerfiles/runtime/Dockerfile.deps-al2023
+++ b/scripts/dockerfiles/runtime/Dockerfile.deps-al2023
@@ -1,0 +1,20 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:2023 as dependencies
+
+# Create sysroot directory and install minimal runtime dependencies
+RUN mkdir /sysroot && \
+    dnf --releasever=$(rpm -q system-release --qf '%{VERSION}') \
+    --installroot /sysroot \
+    -y \
+    --setopt=install_weak_deps=False \
+    install \
+        libyaml \
+        systemd-libs \
+        openssl-libs \
+        cyrus-sasl-lib \
+        libstdc++
+
+# Final minimal runtime image using scratch base
+FROM scratch
+
+# Copy the complete sysroot with all runtime dependencies
+COPY --from=dependencies /sysroot /
--- a/scripts/core_uploader.sh
+++ b/scripts/core_uploader.sh
@@ -34,10 +34,11 @@ export CRASH_FOLDER="/cores-out" # Add all crash files to a cores-out folder for
 export CORE_FILENAME=`basename $S3_KEY_PREFIX`_`date +"%FT%H%M%S"`${HOSTNAME+_host-$HOSTNAME}_${RUN_ID}
 
 # Check if there is a corefile
-export NO_CORE_STR="ls: cannot access $CRASH_FOLDER/core*: No such file or directory"
 export LS_CORE_STR="$(ls $CRASH_FOLDER/core* 2>&1 >/dev/null)"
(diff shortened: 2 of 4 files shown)


scripts/dockerfiles/runtime/Dockerfile.deps-al2023
```diff
@@ -0,0 +1,20 @@
...
+    -y \
+    --setopt=install_weak_deps=False \
+    install \
+        libyaml \
+        systemd-libs \
+        openssl-libs \
+        cyrus-sasl-lib \
+        libstdc++
+
+# Final minimal runtime image using scratch base
```
sparrc, line 16: Doesn't the "final runtime image" need to have the fluent-bit binary copied in?

sparrc: Based on [the doc](https://docs.aws.amazon.com/linux/al2023/ug/barebones-containers.html) you referenced, I was expecting the fluent-bit Dockerfile to look more like this one, in which dependencies and the fluent-bit compilation both happen in a "build" step, and then both the dependencies and the fluent-bit binary are copied into the final image together. (obviously replacing the hello world app with fluent-bit in this example):
```
FROM public.ecr.aws/amazonlinux/amazonlinux:2023 as build
COPY hello-world.c /
RUN dnf -y install gcc
RUN gcc -o hello-world hello-world.c
RUN mkdir /sysroot
RUN mv hello-world /sysroot/
RUN dnf --releasever=$(rpm -q system-release --qf '%{VERSION}') \
  --installroot /sysroot \
  -y \
  --setopt=install_weak_deps=False \
  install glibc && dnf --installroot /sysroot clean all

FROM scratch
COPY --from=build /sysroot /
WORKDIR /
ENTRYPOINT ["/hello-world"]
EOF
```

ShelbyZ: With the AL2 dockerfile rework we pass in a RUNTIME_IMAGE to dockerfiles containing the dependencies for the platform

https://github.com/aws/aws-for-fluent-bit/blob/bd01285f17cc47fb4c45f5b056b47d1488ae6859/Makefile#L77

Then inside the folliow-up dockerfiles we copy in the required binaries/plugins

https://github.com/aws/aws-for-fluent-bit/blob/bd01285f17cc47fb4c45f5b056b47d1488ae6859/scripts/dockerfiles/runtime/Dockerfile#L4-L12

singholt (approved):

danehlim (approved):

PR #997, opened 2025-09-10
Build changes for aws-for-fluent-bit vNext

prateekchaudhry:
### Summary

This PR adds buildspec changes for the upcoming aws-for-fluent-bit 3.x (fluent-bit 4.x) image build. Changes in scope of this PR are related to supporting new image build, while also supporting existing image aws-for-fluent-bit 2.x build. Other buildspec changes will follow which will be related to how the images get tested and published.

## Key changes

* linux.version - This is now a list of versions supported. This is also the main source of truth of which image to build. Properties like `flb-repository`, `major-version`, `al-tag` have also been added to this json which will help in building images
* get_linux_version.sh - Added a file to read the above linux.version file parameters
* build.sh - Moved bulk of buildspec.yml commands to a script build.sh. This script will build and upload pre-publish images to ECR. Keeping tags for existing image same, and for new image adding tags like `latest-3` for uploading to ECR

### Testing

`make debug` succeeded:  yes
Integ tests succeeded:  yes
New tests cover the changes:  no new tests

Successfully built image build in personal codepipeline

Build changes for aws-for-fluent-bit vNext

### Licensing


buildspec.yml
```diff
@@ -17,76 +17,10 @@ phases:
...
-          echo "Building AWS for Fluent Bit version 3 (no-op for now)"
-          echo "Version 3 build logic will be implemented later"
-          # Placeholder for future BUILD_VERSION=3 implementation
-          
-        else
-          echo "Unsupported BUILD_VERSION: $BUILD_VERSION"
-          echo "Supported versions are: 2, 3"
-          exit 1
-        fi
+      # Execute the build script which handles both BUILD_VERSION 2 and 3
```
prateekchaudhry, line 21: `BUILD_VERSION` - set as env var in codepipeline

scripts/publish.sh
```diff
@@ -18,10 +18,19 @@ scripts=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 cd "${scripts}"
 
 IMAGE_SHA_MATCHED="FALSE"
-AWS_FOR_FLUENT_BIT_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION)
+# TODO: Make it work for both build versions (or move entire logic in this script)
+if [ -z "$BUILD_VERSION" ]; then
+    AWS_FOR_FLUENT_BIT_VERSION=$(cat ../AWS_FOR_FLUENT_BIT_VERSION)
```
prateekchaudhry, line 23: Defaulting to current behavior in publish.sh when BUILD_VERSION is not set. This will be removed eventually.

buildspec.yml
```diff
@@ -17,76 +17,10 @@ phases:
...
+      - chmod +x scripts/build.sh
```
ShelbyZ: Why is this required? Do we not need to just set the executable bit when checking in?

linux.version
```diff
@@ -1,11 +1,29 @@
...
+            "flb-repository": "https://github.com/amazon-contributing/upstream-to-fluent-bit.git"
+        }
+    },
+    {
+        "linux": {
+            "major-version": "3",
+            "version": "3.0.0",
+            "latest": "false",
+            "build": "1",
+            "fluent-bit": "v4.0.3",
```
ShelbyZ: nit: Let's pick 4.0.9 until our target becomes available

scripts/build.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
+set -e
+
+# Check BUILD_VERSION environment variable and set default if not provided
+if [ -z "$BUILD_VERSION" ]; then
+    echo "BUILD_VERSION environment variable not provided, defaulting to BUILD_VERSION=2"
+    export BUILD_VERSION=2
+else
+    echo "BUILD_VERSION is set to: $BUILD_VERSION"
```
ShelbyZ: i think this might be excessive as line19 prints the build version if valid

scripts/build.sh
```diff
@@ -0,0 +1,148 @@
...
+    FIREHOSE_PLUGIN_TAG=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "firehose-plugin")
+    CLOUDWATCH_PLUGIN_TAG=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "cloudwatch-plugin")
+    
+    echo "Using KINESIS_PLUGIN_TAG: $KINESIS_PLUGIN_TAG"
+    echo "Using FIREHOSE_PLUGIN_TAG: $FIREHOSE_PLUGIN_TAG"
+    echo "Using CLOUDWATCH_PLUGIN_TAG: $CLOUDWATCH_PLUGIN_TAG"
+    
+    # Export all plugin variables as environment variables for make
+    export KINESIS_PLUGIN_TAG FIREHOSE_PLUGIN_TAG CLOUDWATCH_PLUGIN_TAG
+fi
```
ShelbyZ: Why are these only set for v3 and not v2? Should we not just pull the values in a helper function and call/set it for both?

scripts/build.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Command to build debug image with version-specific parameters
+if [ "$BUILD_VERSION" = "2" ]; then
+    make debug AL_TAG=$AL_TAG FLB_REPOSITORY="$FLB_REPOSITORY" AWS_FOR_FLUENT_BIT_VERSION="$AWS_FOR_FLUENT_BIT_VERSION"
+    make release AL_TAG=$AL_TAG FLB_REPOSITORY="$FLB_REPOSITORY" AWS_FOR_FLUENT_BIT_VERSION="$AWS_FOR_FLUENT_BIT_VERSION"
+else
+    # BUILD_VERSION = 3
+    make debug AL_TAG=$AL_TAG FLB_VERSION=$FLB_VERSION FLB_REPOSITORY="$FLB_REPOSITORY" AWS_FOR_FLUENT_BIT_VERSION="$AWS_FOR_FLUENT_BIT_VERSION" KINESIS_PLUGIN_TAG=$KINESIS_PLUGIN_TAG FIREHOSE_PLUGIN_TAG=$FIREHOSE_PLUGIN_TAG CLOUDWATCH_PLUGIN_TAG=$CLOUDWATCH_PLUGIN_TAG
+    make release AL_TAG=$AL_TAG FLB_VERSION=$FLB_VERSION FLB_REPOSITORY="$FLB_REPOSITORY" AWS_FOR_FLUENT_BIT_VERSION="$AWS_FOR_FLUENT_BIT_VERSION" KINESIS_PLUGIN_TAG=$KINESIS_PLUGIN_TAG FIREHOSE_PLUGIN_TAG=$FIREHOSE_PLUGIN_TAG CLOUDWATCH_PLUGIN_TAG=$CLOUDWATCH_PLUGIN_TAG
+fi
```
ShelbyZ: If we always set/extract plugin versions we would not need this if/else and just set the values and call make release/debug

scripts/build.sh
```diff
@@ -0,0 +1,148 @@
...
+# Push the image to ECR with corresponding architecture as the tag.
+aws ecr get-login-password --region ${AWS_REGION}| docker login --username AWS --password-stdin ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com
+aws ecr create-repository --repository-name amazon/aws-for-fluent-bit-test --image-scanning-configuration scanOnPush=true --region ${AWS_REGION}  || true
+
+# Get architecture and set image tag suffix based on BUILD_VERSION
+# TODO: Simplify logic, just use AL_TAG
+if [ "$BUILD_VERSION" = "2" ]; then
+    IMAGE_TAG_SUFFIX="al2"
+else
+    IMAGE_TAG_SUFFIX=al"$AL_TAG"
```
ShelbyZ: I would expect both `IMAGE_TAG_SUFFIX=al${AL_TAG}` without needing to hardcode v2

scripts/build.sh
```diff
@@ -0,0 +1,148 @@
...
+    MANIFEST_INIT_TAG="init-latest-$BUILD_VERSION"
+    ARM64_TAG="arm64-$BUILD_VERSION"
+    AMD64_TAG="amd64-$BUILD_VERSION"
+    INIT_ARM64_TAG="init-arm64-$BUILD_VERSION"
+    INIT_AMD64_TAG="init-amd64-$BUILD_VERSION"
+fi
+
+docker manifest create ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$MANIFEST_LATEST_TAG ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$ARM64_TAG ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$AMD64_TAG || true
+docker manifest annotate --arch arm64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$MANIFEST_LATEST_TAG ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$ARM64_TAG || true
+docker manifest annotate --arch amd64 ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$MANIFEST_LATEST_TAG ${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test:$AMD64_TAG || true
```
ShelbyZ: It might be worth putting this into a helper function that takes:
1. Manifest
2. arm64 image
3. amd64 image

Then have it create manifest, annotate images, inspect, and push. That would allow reuse across the image s handled below

scripts/get_linux_version.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ShelbyZ: 2025

scripts/build.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
```
ShelbyZ, line 1: Maybe add a header block similar to other script

prateekchaudhry: Thanks for review, I will address the comments in a revision

scripts/build.sh
prateekchaudhry: I think we can do for both here. I did not want to make any changes to v2 build recipe earlier, and let build_plugins.sh set it. But now that I have helper function set, I  think it can be quite safe to abstract out everything here. I will give it a try

buildspec.yml
prateekchaudhry: addressed in new commit

ShelbyZ (approved):

Makefile
```diff
@@ -22,7 +22,7 @@ FLB_VERSION ?= "1.9.10"
 # Fluent Bit repository to checkout, will use value if not set
 FLB_REPOSITORY ?= "https://github.com/amazon-contributing/upstream-to-fluent-bit.git"
 # AWS for Fluent Bit Version
-AWS_FOR_FLUENT_BIT_VERSION ?= $(shell cat AWS_FOR_FLUENT_BIT_VERSION)
+AWS_FOR_FLUENT_BIT_VERSION ?= $(shell ./scripts/get_linux_version.sh 2 version)
```
singholt, line 25: this needs jq installed right? 

1. Is that fulfilled by all the makefile targets using this
2. any other pre-reqs?

scripts/build.sh
```diff
@@ -0,0 +1,140 @@
...
+set -e
+
+# Validate BUILD_VERSION variable is set
+if [ "$BUILD_VERSION" != "2" ] && [ "$BUILD_VERSION" != "3" ]; then
+    echo "Unsupported BUILD_VERSION: $BUILD_VERSION"
+    echo "Supported versions are: 2, 3"
+    exit 1
+fi
+
+echo "Building AWS for Fluent Bit version $BUILD_VERSION"
```
singholt, line 23: Is this script starting from this line and below a copy paste? Just making sure

scripts/publish.sh
```diff
@@ -603,8 +612,6 @@ verify_ecr() {
 check_image_version() {
 	export DOCKER_CLI_EXPERIMENTAL=enabled
 	EXIT_CODE=0
-
-	docker_hub_login
```
singholt, line 607: Whats the context here?

linux.version
```diff
@@ -1,11 +1,29 @@
...
+            "cloudwatch-plugin": "v1.9.4",
+            "al-tag": "2",
+            "flb-repository": "https://github.com/amazon-contributing/upstream-to-fluent-bit.git"
+        }
+    },
+    {
+        "linux": {
+            "major-version": "3",
+            "version": "3.0.0",
+            "latest": "false",
```
singholt, line 19: nit: if I understand correctly, this is a flag to denote whether to publish latest tag or not, rather than to denote whether this image version is the latest of all available versions.

I would suggest to rename this to `publish-latest-image-tag` or similar

Makefile
prateekchaudhry: Yes this would need jq installed. Checking codebuild repos and it should be there in codebuild env ([example](https://github.com/aws/aws-codebuild-docker-images/blob/1e24f33651ce49697c4e413b8bf2f2f1dae4df99/unsupported_images/al2/x86_64/standard/2.0/Dockerfile#L32)). We also seem to be using jq already in [build_plugins.sh](https://github.com/aws/aws-for-fluent-bit/blob/0173ad9187239926f69b67f5953ff29be0268fe8/scripts/build_plugins.sh#L133) following the make target code path.

scripts/publish.sh
prateekchaudhry: docker hub login is not needed here, so I removed it

linux.version
prateekchaudhry: TODO in a later PR

scripts/build.sh
prateekchaudhry: Yes functionally this is same as existing buildspec.yml. Moved code to this script and refactored

Makefile
prateekchaudhry: TODO in later pr : Update readme that jq needed in local build

singholt (approved):

PR #1005, opened 2025-09-24
Add helper script to support syncing all test images

ShelbyZ:
### Summary
- Remove duplicate test repo sync code
- Add helper script for sync/tag images
- Update buildspec's usage

### Testing
Local script testing against test repo to see images synced and retag. Version 3 images (-3) are not synced from the test repo as they are not found.

Enhancement: Update buildspec to sync version 3 images from test repo when found

### Licensing


scripts/pipeline/sync-test-images.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# Set up repository base URL
+REPO_BASE="${AWS_ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/amazon/aws-for-fluent-bit-test"
+
+# Define image tags to pull
+STANDARD_TAGS=("amd64" "arm64" "amd64-debug" "arm64-debug")
+INIT_TAGS=("init-amd64" "init-arm64" "init-amd64-debug" "init-arm64-debug")
+
+# Helper function to pull and optionally retag a single image
+pull_and_retag_image() {
```
mye956, line 21: nit: could we move these helper functions towards the beginning of the file before all of the standalone commands that may use them (makes it also a bit easier to read the file)? ditto to also consider grouping the standalone commands in a `main` and/or other helper functions as see you see fit.

scripts/pipeline/sync-test-images.sh
```diff
@@ -0,0 +1,135 @@
...
+        # Retag if --tag flag was provided
+        if [[ "$RETAG_IMAGES" == "true" ]]; then
+            local new_tag="amazon/aws-for-fluent-bit:${full_tag}"
+            echo "Retagging $full_image -> $new_tag"
+            docker tag "$full_image" "$new_tag"
+        fi
+        return 0
+    else
+        # Pull failed
+        if [[ "$allow_failure" == "true" ]]; then
```
mye956, line 41: Q: just curious, if the idea is we want to keep track of which tags failed to be pulled what does the "exit" condition help with?

scripts/pipeline/sync-test-images.sh
```diff
@@ -0,0 +1,135 @@
...
+    echo "Pulling init process images${suffix:+ with suffix \"$suffix\"}..."
+    for tag in "${INIT_TAGS[@]}"; do
+        if ! pull_and_retag_image "$tag" "$suffix" "$allow_failure"; then
+            failed_images+=("${tag}${suffix}")
+        fi
+    done
+    
+    if [[ ${#failed_images[@]} -gt 0 && "$allow_failure" == "true" ]]; then
+        echo "Warning: The following images with suffix \"$suffix\" failed to pull:"
+        printf '  %s\n' "${failed_images[@]}"
```
mye956, line 73: Q: why do we not want to exit here?

scripts/pipeline/sync-test-images.sh
```diff
@@ -0,0 +1,135 @@
...
+pull_image_set "" "false"
+
+# Pull images with -3 suffix (failures are warnings)
+pull_image_set "-3" "true"
+
+echo "Successfully synced test repo images to local Docker"
+
+if [[ "$RETAG_IMAGES" == "true" ]]; then
+    echo "Images have been retagged as: amazon/aws-for-fluent-bit"
+    echo ""
```
mye956: nit(non-blocking): could just consider using `printf` here instead for new lines. ditto on L.132-133

scripts/pipeline/sync-test-images.sh
ShelbyZ: For the v2 images (lack suffix) we expect them to exist, so a pull failure should bubble up and fail the task. For the v3 images we want to track it as a warning until they are available. A follow-up PR post release will remove these allowed failures

scripts/pipeline/sync-test-images.sh
ShelbyZ: Similar to above, for v3 images they may not be available until we do the first publish. If we enable the failure now then we will have issues with publishing v2 images

scripts/pipeline/sync-test-images.sh
mye956: gotcha, thanks for the context. might want to add a note somewhere in the PR for reference 😅

scripts/pipeline/sync-test-images.sh
mye956: context was given above

scripts/pipeline/sync-test-images.sh
mye956: this can be address in a follow up

mye956 (approved):

prateekchaudhry (approved):

PR #1007, opened 2025-09-25
Add BUILD_VERSION=3 support to publish_dockerhub

ShelbyZ:
### Summary
- Add BUILD_VERSION=3 checks to override image suffix
- Verify images exist for BUILD_VERSION=3 or skip
- Add BUILD_VERSION=3 checks to dockerhub verification and skip if images do not exist

### Testing
Local testing/debugging, changes follow similar pattern to allow BUILD_VERSION=3 to gracefully fail/warn and not prevent the existing action from failing.

Enhancement: Add support for publishing BUILD_VERSION=3 images to Dockerhub

### Licensing

### DIFF
--- a/buildspec_publish_dockerhub.yml
+++ b/buildspec_publish_dockerhub.yml
@@ -13,10 +13,12 @@ phases:
       - ./scripts/pipeline/sync-test-images.sh --tag
 
       # Push the image to dockerhub
-      - 'DRY_RUN="false" ./scripts/publish.sh cicd-publish dockerhub'
+      - './scripts/publish.sh cicd-publish dockerhub'
+      - BUILD_VERSION=3 ./scripts/publish.sh cicd-publish dockerhub
 
       # Pull the image from dockerhub and verify
       - './scripts/publish.sh cicd-verify dockerhub'
+      - BUILD_VERSION=3 ./scripts/publish.sh cicd-verify dockerhub
 artifacts:
   files:
     - '**/*'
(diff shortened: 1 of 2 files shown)


prateekchaudhry (approved):

buildspec_publish_dockerhub.yml
```diff
@@ -13,10 +13,12 @@ phases:
       - ./scripts/pipeline/sync-test-images.sh --tag
 
       # Push the image to dockerhub
-      - 'DRY_RUN="false" ./scripts/publish.sh cicd-publish dockerhub'
```
TheanLim, line 16: Is `DRY_RUN` not being used anymore?

scripts/publish.sh
TheanLim, line 844: Just for my own knowledge: when do we use `publish dockerhub` vs `cicd-publish dockerhub`

TheanLim, line 916: For my own knowledge: when do we use `verify dockerhub` vs `cicd-verify dockerhub`?
It seems to perform simpler verification compared to using `verify_dockerhub()`

buildspec_publish_dockerhub.yml
ShelbyZ: It has not been available for awhile, so figured we should remove it

TheanLim (approved):

scripts/publish.sh
ShelbyZ: I think this is a tech-debt question. We use `cicd-publish dockerhub` from the build pipeline and for the other I am less sure. We need to refactor/simplify these scripts post-launch

ShelbyZ: From the buildspec for cicd-verify dockerhub and the other is likely a lost option (tech-debt)

PR #1025, opened 2025-10-17
Add Package Versions/Adding Packages to VERSIONS.md

ShelbyZ:
### Summary
- Add details on how to query installed packages from Public ECR
- Add details on how to add use-case specific packages to a custom image

### Testing
Tested commands locally and built custom images

### Licensing

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -50,6 +50,10 @@ Example Labels
 }
 ```
 
+## Linux Package Versions
+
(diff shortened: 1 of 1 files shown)


VERSIONS.md
```diff
@@ -50,6 +50,10 @@ Example Labels
 }
 ```
 
+## Linux Package Versions
+
+To get a full listing of installed package versions, run: `docker run public.ecr.aws/aws-observability/aws-for-fluent-bit:3.0.0 grep "Installed:" /var/log/dnf.log /var/log/yum.log 2>/dev/null | sed 's/.*Installed: //'`. This can be useful to determine if a specific version is affected by a CVE.
+
 ## Version 3.0.0
```
prateekchaudhry, line 57: non blocking - Maybe we should be calling it Version 3.x here, as we bump up patch and build/OS versions

prateekchaudhry (approved):

mye956 (approved):

PR #1044, opened 2025-11-21
Fix long build times for AL2023 by setting nofile limit

ShelbyZ:
### Summary
Long build times on AL2023 image (4-6 hours) seem to be related to the default NOFile limit set to infinity. By setting a distinct limit in the dockerfile during deps image build this is not an issue - https://github.com/docker/buildx/issues/379

### Testing
`make release` and `make debug` around normal 5-10min system depending...

### Licensing

### DIFF
--- a/scripts/dockerfiles/runtime/Dockerfile.deps-al2023
+++ b/scripts/dockerfiles/runtime/Dockerfile.deps-al2023
@@ -3,6 +3,8 @@ FROM public.ecr.aws/amazonlinux/amazonlinux@${OS_DIGEST} as dependencies
 
 # Create sysroot directory and install minimal runtime dependencies
 RUN mkdir /sysroot && \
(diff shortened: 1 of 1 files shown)


harishxr (approved):

prateekchaudhry (approved):

prateekchaudhry:
(non blocking but curious about this) - `Long build times on AL2023 image (4-6 hours)` - is the long build time only AL2023 recipe specific? Does AL2 build need an update too?

Thanks

ShelbyZ:
> (non blocking but curious about this) - `Long build times on AL2023 image (4-6 hours)` - is the long build time only AL2023 recipe specific? Does AL2 build need an update too?
> 
> Thanks

I only observed this on when building on AL2023 based system and specifically when building the AL2023 deps images.

PR #1074, opened 2026-02-17
fix: Update check-update for os-digest and publish flag

ShelbyZ:
### Summary

- only stage changes if at least one version publishing
- remove amazon-linux-sha, use os-digest
- remove todo from github action

fix: Update check-update for os-digest and publish flag

### Licensing

### DIFF
--- a/linux.version
+++ b/linux.version
@@ -3,7 +3,7 @@
     "linux": {
       "major-version": "2",
       "version": "2.34.3.20260209",
-      "release-version": "2.34.1.20251031",
+      "release-version": "2.34.3.20260209",
       "latest": "true",
       "build": "1",
       "fluent-bit": "1.9.10",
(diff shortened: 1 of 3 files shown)


mye956 (approved):

linux.version
```diff
@@ -3,7 +3,7 @@
     "linux": {
       "major-version": "2",
       "version": "2.34.3.20260209",
-      "release-version": "2.34.1.20251031",
+      "release-version": "2.34.3.20260209",
```
singholt, line 6: curious - whats the difference between version and release-version?

singholt (approved):

linux.version
ShelbyZ: A way to track what version is releasing, we would need to be an extensive introspection piece to allow the github action to see that a release was currently in progress to prevent initiate-release from triggering if we tried to track fluent-bit version via the public image (and AL sha via os-digest). Can talk more offline (or ask Mike :))

PR #433, opened 2022-09-27
Bump test loadtests 4

matthewfala:
Added termination protection on instances protecting load tests from transient host kill errors

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1082, opened 2026-02-27
enhancement: adjust image tagging during release

ShelbyZ:
### Summary
- Use version field from linux.version to tag v2/v3 images in -test ECR repository
- adjust tagging in scripts/buildspecs
- Add -qq to unzip commands to remove thousands of log lines
- Remove scale-in/termination protection on load test resources as it sometimes fails cleanup

### Testing
Local testing with personal codebuild pipeline up to image release. The image retagging to specific versions is working for integ/load tests and should also work for ECR/DockerHub

enhancement: adjust image tagging during release

### Licensing

### DIFF
--- a/load_tests/create_testing_resources/ecs/app.py
+++ b/load_tests/create_testing_resources/ecs/app.py
@@ -50,12 +50,13 @@ class TestingResources(Stack):
             desired_capacity=5,
             vpc=vpc,
             vpc_subnets={ 'subnet_type': ec2.SubnetType.PUBLIC },
+            new_instances_protected_from_scale_in=False
         )
         asg.apply_removal_policy(RemovalPolicy.DESTROY)
 
         capacity_provider = ecs.AsgCapacityProvider(self, "asgCapacityProvider",
             auto_scaling_group=asg,
-            enable_managed_termination_protection=True
+            enable_managed_termination_protection=False
         )
         cluster.add_asg_capacity_provider(capacity_provider)
         cluster.apply_removal_policy(RemovalPolicy.DESTROY)
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -39,6 +39,7 @@ AWS_FOR_FLUENT_BIT_VERSION=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "ve
 KINESIS_PLUGIN_TAG=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "kinesis-plugin")
 FIREHOSE_PLUGIN_TAG=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "firehose-plugin")
 CLOUDWATCH_PLUGIN_TAG=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "cloudwatch-plugin")
+VERSION=$(./scripts/get_linux_version.sh "$BUILD_VERSION" "version")
 
 IMAGE_TAG_SUFFIX=al"$AL_TAG"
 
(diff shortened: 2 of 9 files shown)


mye956 (dismissed):

scripts/build.sh
```diff
@@ -125,36 +118,18 @@ tag_push_and_scan "amazon/aws-for-fluent-bit:init-debug-$IMAGE_TAG_SUFFIX" "$INI
...
-    ARM64_TAG="arm64-$BUILD_VERSION"
-    AMD64_TAG="amd64-$BUILD_VERSION"
-    INIT_ARM64_TAG="init-arm64-$BUILD_VERSION"
-    INIT_AMD64_TAG="init-amd64-$BUILD_VERSION"
-    DEBUG_ARM64_TAG="arm64-debug-$BUILD_VERSION"
-    DEBUG_AMD64_TAG="amd64-debug-$BUILD_VERSION"
-    INIT_DEBUG_ARM64_TAG="init-arm64-debug-$BUILD_VERSION"
-    INIT_DEBUG_AMD64_TAG="init-amd64-debug-$BUILD_VERSION"
-fi
+MANIFEST_LATEST_TAG="$VERSION"
```
mye956, line 121: q: we originally had `latest-$VERSION` as part of this change before. just curious why we decided to opt for just the version as the manifest tag?

ShelbyZ: When checking the merge conflicts, I noticed we had it and wondered if it added any value to have `latest-$VERSION`. We do not have a `latest` for v3 images and no public facing versioned latest.

I made the choice to remove it from this testing repo and update the references to just `$VERSION` to be more consistent with what the public facing repositories expose.

load_tests/create_testing_resources/ecs/app.py
```diff
@@ -50,12 +50,13 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
             vpc=vpc,
             vpc_subnets={ 'subnet_type': ec2.SubnetType.PUBLIC },
+            new_instances_protected_from_scale_in=False
         )
         asg.apply_removal_policy(RemovalPolicy.DESTROY)
 
         capacity_provider = ecs.AsgCapacityProvider(self, "asgCapacityProvider",
             auto_scaling_group=asg,
-            enable_managed_termination_protection=True
+            enable_managed_termination_protection=False
```
TheanLim, line 59: Is this still the case? https://github.com/aws/aws-for-fluent-bit/pull/433

> Added termination protection on instances protecting load tests from transient host kill errors

prateekchaudhry (approved):

load_tests/create_testing_resources/ecs/app.py
ShelbyZ: The load-tests work as:
1. Create resources (VPC, EC2 hosts, ASG, etc)
2. Run tasks
3. Cleanup 1

Where we see failures is step 3 that has transient failures on having the tasks stopped so the scale-in on the ASG can trigger the EC2 host termination. We do not expect any transient scale-in events that would cause instance termination prematurely here.

https://repost.aws/knowledge-center/auto-scaling-scale-down

I could also opt to remove it as it was unrelated (annoying during testing of these changes)

scripts/build.sh
ShelbyZ: Synced offline and we should be able to keep the change by updating tag lookup

sparrc (approved):

mye956 (approved):

PR #1099, opened 2026-04-06
Fix image tags for dockerhub/public ecr

ShelbyZ:
### Summary
Image tag lookup for dockerhub/public ECR was using older `-3` checks and not the new versioned images which caused them to get missed from the last release.

This change also removes the older version 3 safe guards put in places for the first release - safe to remove now.

### Licensing

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -154,31 +154,18 @@ publish_to_docker_hub() {
 			create_manifest_list ${1} "stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
 		fi
 	else
-		source_suffix=""
-		# Add image suffix for BUILD_VERSION=3
-		if [ "$BUILD_VERSION" = "3" ]; then
-			source_suffix="-3"
-
-			# Verify all images exist before proceeding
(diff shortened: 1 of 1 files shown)


scripts/publish.sh
```diff
@@ -154,31 +154,18 @@ publish_to_docker_hub() {
 			create_manifest_list ${1} "stable" ${AWS_FOR_FLUENT_BIT_STABLE_VERSION}
 		fi
 	else
-		source_suffix=""
-		# Add image suffix for BUILD_VERSION=3
-		if [ "$BUILD_VERSION" = "3" ]; then
-			source_suffix="-3"
-
-			# Verify all images exist before proceeding
-			if ! verify_images "-3"; then
```
mye956, line 163: could be done in a follow up but it might be a good idea to verify the images either before and/or after publishing

mye956 (approved):

scripts/publish.sh
ShelbyZ: This was previously added to prevent the v3 images from early release at various stages. It may be worth considering during a script rework.

harishxr (approved):

PR #1102, opened 2026-04-07
enhancement: temporarily check and change scanning type to BASIC before building new images

mye956:
### Summary
This PR first checks if the scanning type of our ECR repos are `BASIC` and if it is then we change the typing. This is a short term workaround as our image builder script isn't compatible with `ENHANCED` typing.

*Issue #, if available:*

### Testing

`make debug` succeeded: 
Integ tests succeeded: 
New tests cover the changes: 

* enhancement: temporarily check and change scanning type to BASIC before building new images

### Licensing

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -6,6 +6,16 @@ phases:
       python: 3.x
   pre_build:
     commands:
+      # Short term fix: ensure ECR basic scanning is enabled for the registry
+      # TODO: Remove this once we have a long term fix
+      - |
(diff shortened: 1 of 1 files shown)


ShelbyZ (approved):

buildspec.yml
```diff
@@ -6,6 +6,16 @@ phases:
       python: 3.x
   pre_build:
     commands:
+      # Short term fix: ensure ECR basic scanning is enabled for the registry
+      # TODO: Remove this once we have a long term fix
```
TheanLim, line 10: Just curious, do we have a rough idea on how the long term fix would look like?

TheanLim (approved):

buildspec.yml
mye956: either root cause what/why is the scanning type changing to `ENHANCED` or have a dedicated standalone workflow that does this separately from the image builder CodeBuild project

PR #1115, opened 2026-04-20
enhancement: rework core_uploader/debug images

ShelbyZ:
### Summary

- rewrite core_uploader as a supervisor for debugging (core, heap, rss)
- add support for SIGCONT, SIGUSR1 (gcore snapshot), SIGUSR2 (heap snapshot)
- profile rss during run
- profile jemalloc heap during run (needs env value setup, see docs)
- remove redundant EFS images into unified debug
- rework init-debug to use core_uploader
- add libunwind (required for gcore snapshot)
- compile jemalloc profiling support for debug builds
- add docs (debug/heap-profiling)
- update docs (valgrind, debugging, READMEs)

### Why this change

The current debug image focuses on running and uploading a core file on crash. This is useful unless the application does not crash or no core file is created. The valgrind image uses valgrind for analysis for leaks, but is not useful for debugging memory growth during application lifetime.

This rework aims to enable tracking heap/rss and use the same upload mechanisms to capture on exit and midrun. 

Using the jemalloc heaps and rss we can track allocations and discover where memory may be lingering during run. This helps track growth and suggest changes to fluent-bit in the case the memory is not leaked (on exit).

### Testing
- Built AL2/AL2023 debug/init-debug variants
- Ran each image to check that core files, heap files, and rss files were produced
- tested EFS wait duration and potential upload

enhancement: rework core_uploader/debug images

### Licensing


scripts/core_uploader.sh
```diff
@@ -1,96 +1,295 @@
...
+
+# Block until fluent-bit exits, re-waiting after signal interrupts, then collect artifacts.
+wait_for_shutdown() {
+    log "Fluent Bit running (PID $FLB_PID), waiting for exit"
+
+    while kill -0 "$FLB_PID" 2>/dev/null; do
+        wait "$FLB_PID" 2>/dev/null
+    done
+
+    log "Fluent Bit exited (exit code: $?)"
```
singholt: if kill -0 ... fails, then $? after the while loop always reflects the exit status of the failed kill -0 (which is 1), not fluent-bit's actual exit code.

```suggestion
    while kill -0 "$FLB_PID" 2>/dev/null; do
        wait "$FLB_PID" 2>/dev/null
        flb_exit=$?
    done
    
    log "Fluent Bit exited (exit code: $flb_exit)"
```

scripts/core_uploader.sh
```diff
@@ -1,96 +1,295 @@
 #!/bin/bash
```
singholt, line 1: nit: missing copyright on this file and valgrind_uploader.sh

ShelbyZ: I'll bring it back (and submit some follow-ups to add headers across all files

scripts/core_uploader.sh
ShelbyZ: That is a good catch, we could see 1 (failure) or 64 (partial success). Let me see about an update

ShelbyZ: Reworked to better capture a failed kill exit (log) and attempt to report the fluent-bit exit

singholt (approved):

prateekchaudhry (approved):

PR #1178, opened 2026-07-20
fix: use merge time for changelog release boundary

TheanLim:
## What

`get_changes_since_last_release` in `scripts/generate_changelog.sh` computed the "last release" boundary with:

```
--search "head:release- sort:updated-desc" --limit 1
```

`sort:updated-desc` orders by last-touched time, not merge time. An old release PR whose branch was later deleted or cross-referenced floats to the top and sets the boundary to its (stale) `mergedAt`. In the 20260720 release this selected PR #979 (merged 2025-08-13, updated 2026-07-13), pushing the boundary ~11 months too early and sweeping already-shipped PRs (#1107–#1173) into the new entry.

## Fix

GitHub search has no `sort:merged`, so fetch a window of release PRs with `sort:updated-desc` and pick the max `mergedAt` client-side. Also tighten `merged:>=` to `merged:>` so the boundary release PR itself is excluded.

## Testing

Ran the script against the 20260720 release state:
- Before: boundary `2025-08-13`, 3.4.9 entry had 19 bullets (#1107–#1176).
- After: boundary `2026-07-13T20:10:43Z` (#1175), 3.4.9 entry has 2 bullets (AL package line + #1176) — the only PR merged since the last release.

The corresponding CHANGELOG.md correction for the 20260720 release is in PR #1177.

### DIFF
--- a/scripts/generate_changelog.sh
+++ b/scripts/generate_changelog.sh
@@ -60,8 +60,7 @@ get_changes_since_last_release() {
 		--limit 30 \
 		--json mergedAt,headRefName \
 		--jq '[.[]
-			| select(.headRefName | test("^release-"))
-			| select(.mergedAt != null)]
+			| select(.headRefName | test("^release-"))]
 			| sort_by(.mergedAt) | last | .mergedAt' 2>/dev/null || true)
 
 	if [ -z "$last_release_merged_at" ] || [ "$last_release_merged_at" = "null" ]; then
--- a/scripts/generate_changelog.sh
+++ b/scripts/generate_changelog.sh
@@ -47,14 +47,22 @@ get_version_info() {
 # Get PR titles merged since the last "Release XXXX" commit via the GitHub CLI (gh) to query merged PRs by date
 get_changes_since_last_release() {
 	local last_release_merged_at
+	# Boundary = merge time of the most-recently-merged release PR.
+	# GitHub search cannot sort by merge time (no sort:merged), and sort:updated
+	# floats stale release PRs (old branches deleted/cross-referenced long after
+	# they merged) to the top, producing a far-too-early boundary. Fetch a window
+	# of release PRs and pick the max mergedAt client-side.
 	last_release_merged_at=$(gh pr list \
 		--repo "$REPO" \
 		--state merged \
 		--base mainline \
-		--search "head:release- sort:updated-desc" \
-		--limit 1 \
-		--json mergedAt \
-		--jq '.[0].mergedAt' 2>/dev/null || true)
+		--search "head:release-" \
+		--limit 30 \
+		--json mergedAt,headRefName \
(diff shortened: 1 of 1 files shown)


scripts/generate_changelog.sh
```diff
@@ -47,14 +47,21 @@ get_version_info() {
...
+	# of release PRs and pick the max mergedAt client-side.
 	last_release_merged_at=$(gh pr list \
 		--repo "$REPO" \
 		--state merged \
 		--base mainline \
-		--search "head:release- sort:updated-desc" \
-		--limit 1 \
-		--json mergedAt \
-		--jq '.[0].mergedAt' 2>/dev/null || true)
+		--search "head:release-" \
```
TheanLim: this can match any branch with `release` in its name like `hotfix-release-x` so added another check at L63 for the exact prefix match

Yiyuanzzz (dismissed):

scripts/generate_changelog.sh
```diff
@@ -47,14 +47,21 @@ get_version_info() {
...
+		--limit 30 \
```
mye956, line 62: Q: does this mean at most we will only get 30 merged PRs since the last release PR? is this because ofsome kind of limitation on github's side?

TheanLim: No, the 30 only applies to the release PRs we sample to find the boundary. The one that applies to changelog PR is at L75.

Ideally we'd search `sort:merged` + `--limit 1` to grab the latest-merged release PR directly. But GitHub search has no `sort:merged`, so we can't guarantee that PR lands at position 0 — meaning `--limit 1` could return the wrong one. As a workaround we fetch a window of 30 recent release PRs and pick the max `mergedAt` client-side.

30 isn't a hard guarantee — it returns the 30 most-recent merged release PRs by GitHub's order, which is unclear what sorting mechanism is in play. Thinking more of it now, i should add `sort:updated-desc` back. Combining that with the 30 PR window, and `max(mergedAt)` client-side should be the better approach

TheanLim: Or another approach is to not use `sort:updated-desc` but filter down `release` PRs merged in the past year
```
    lookback=$(date -u -d '365 days ago' +%Y-%m-%d 2>/dev/null || date -u -v-365d +%Y-%m-%d)
	last_release_merged_at=$(gh pr list \
		--repo "$REPO" \
		--state merged \
		--base mainline \
		--search "head:release- merged:>=${lookback}" \
		--limit 200 \
		--json mergedAt,headRefName \
		--jq '[.[]
			| select(.headRefName | test("^release-"))]
			| sort_by(.mergedAt) | last | .mergedAt' 2>/dev/null || true)
```

mye956 (approved):

Yiyuanzzz (approved):

PR #1193, opened 2026-08-17
Release 20260817

mye956:
AWS Fluentbit Release Kickoff

PR #1195, opened 2026-08-18
Revert "Release 20260817 (#1193)"

amogh09:
### Summary

This reverts commit 6ef0c09138ec1f64b618b279948223e954505dfd.

Fluent Bit v5.1.0 introduced a config-parse regression so reverting the bump. This commit sets the branch back to the state before #1192. Diff can be seen [here](https://github.com/aws/aws-for-fluent-bit/compare/4656228...revert-1193-fluentbit-v510).

### Licensing

### DIFF
--- a/linux.version
+++ b/linux.version
@@ -14,17 +14,17 @@
       "al-tag": "2",
       "os-digest": "sha256:3ad6628ec681ddbac364a2c4ff87b81082982eb3d7f7031ced9e6268bfab6784",
       "flb-repository": "https://github.com/amazon-contributing/upstream-to-fluent-bit.git",
-      "publish": "false"
+      "publish": "true"
     }
   },
   {
     "linux": {
       "major-version": "3",
-      "version": "3.5.0",
-      "release-version": "3.5.0",
+      "version": "3.4.13",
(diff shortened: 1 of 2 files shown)


linux.version
```diff
@@ -14,17 +14,17 @@
...
       "major-version": "3",
-      "version": "3.5.0",
-      "release-version": "3.5.0",
+      "version": "3.4.13",
+      "release-version": "3.4.13",
       "latest": "false",
       "build": "1",
-      "fluent-bit": "v5.1.0",
+      "fluent-bit": "v5.0.9",
       "release-fluent-bit": "v5.1.0",
```
singholt: I think `release-fluent-bit` also needs to be reverted. its confusing but I suspect the GH action takes this as the source for "fluent-bit"

amogh09: Thank you for catching. Fixed it.

mye956: +1 yes, it's mainly for keeping track of what fluent bit version we have that's currently out there versus what we would like to release

mye956 (approved):

singholt (approved):

